PR #8, opened 2021-06-21
added target parameter

Atticus1806:
This adds the target parameter to Layers in which explicitly setting the parameter is required. For now this is just the `BaseChoice` Layer. Not all layers should have the possibility to set targets, since it does not make sense in most cases. Thats why the same logic as for `n_out` is used.

### DIFF
--- a/models/_generate_layers.py
+++ b/models/_generate_layers.py
@@ -22,7 +22,6 @@ from returnn.tf.layers.basic import ConstantLayer, VariableLayer, CondLayer, Swi
 from returnn.tf.layers.rec import RecLayer, RnnCellLayer, MaskedComputationLayer
 from returnn.tf.layers.rec import PositionalEncodingLayer, RelativePositionalEncodingLayer
 from returnn.tf.layers.rec import BaseChoiceLayer
-from returnn_common.models.base import LayerRef
 _my_dir = os.path.dirname(os.path.abspath(__file__))
 _out_filename = f"{_my_dir}/_generated_layers.py"
 
--- a/models/_generate_layers.py
+++ b/models/_generate_layers.py
@@ -272,7 +272,7 @@ class LayerSignature:
         inspect.Parameter(
           name="target",
           kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
-        param_type_s=LayerRef,
+        param_type_s="LayerRef",
         docstring="explicit target")
     for name, param in self.inspect_init_sig.parameters.items():
       # Ignore a number of params which are handled explicitly.
--- a/models/_generate_layers.py
+++ b/models/_generate_layers.py
@@ -271,7 +271,7 @@ class LayerSignature:
         self,
         inspect.Parameter(
           name="target",
-          kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
+          kind=inspect.Parameter.KEYWORD_ONLY),
         param_type_s="LayerRef",
         docstring="explicit target")
     for name, param in self.inspect_init_sig.parameters.items():
--- a/models/_generate_layers.py
+++ b/models/_generate_layers.py
@@ -253,7 +253,7 @@ class LayerSignature:
     LinearLayer, ConvLayer, TransposedConvLayer, RecLayer, RnnCellLayer,
     PositionalEncodingLayer, RelativePositionalEncodingLayer}
 
-  _LayerClassesWithExplicitTarget={
+  _LayerClassesWithExplicitTarget = {
     BaseChoiceLayer}
 
   def _init_args(self):
--- a/models/_generate_layers.py
+++ b/models/_generate_layers.py
@@ -21,7 +21,8 @@ from returnn.tf.layers.basic import LinearLayer, ConvLayer, TransposedConvLayer
 from returnn.tf.layers.basic import ConstantLayer, VariableLayer, CondLayer, SwitchLayer, SubnetworkLayer
 from returnn.tf.layers.rec import RecLayer, RnnCellLayer, MaskedComputationLayer
 from returnn.tf.layers.rec import PositionalEncodingLayer, RelativePositionalEncodingLayer
-
+from returnn.tf.layers.rec import BaseChoiceLayer
+from returnn_common.models.base import LayerRef
 _my_dir = os.path.dirname(os.path.abspath(__file__))
 _out_filename = f"{_my_dir}/_generated_layers.py"
 
@@ -253,6 +254,9 @@ class LayerSignature:
     LinearLayer, ConvLayer, TransposedConvLayer, RecLayer, RnnCellLayer,
     PositionalEncodingLayer, RelativePositionalEncodingLayer}
 
+  _LayerClassesWithExplicitTarget={
+    BaseChoiceLayer}
+
   def _init_args(self):
     # n_out is handled specially
     if self.layer_class in self._LayerClassesWithExplicitDim:
@@ -263,6 +267,14 @@ class LayerSignature:
           kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
         param_type_s="int",
         docstring="output dimension")
+    if self.layer_class in self._LayerClassesWithExplicitTarget:
+      self.params["target"] = LayerSignature.Param(
+        self,
+        inspect.Parameter(
+          name="target",
+          kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
+        param_type_s=LayerRef,
+        docstring="explicit target")
     for name, param in self.inspect_init_sig.parameters.items():
       # Ignore a number of params which are handled explicitly.
       if param.kind in (param.VAR_POSITIONAL, param.VAR_KEYWORD):


models/_generate_layers.py
```diff
@@ -21,7 +21,8 @@
 from returnn.tf.layers.basic import ConstantLayer, VariableLayer, CondLayer, SwitchLayer, SubnetworkLayer
 from returnn.tf.layers.rec import RecLayer, RnnCellLayer, MaskedComputationLayer
 from returnn.tf.layers.rec import PositionalEncodingLayer, RelativePositionalEncodingLayer
-
+from returnn.tf.layers.rec import BaseChoiceLayer
+from returnn_common.models.base import LayerRef
```
albertz, line 25: There should be one empty line after the imports.

models/_generate_layers.py
```diff
@@ -263,6 +267,14 @@ def _init_args(self):
           kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
         param_type_s="int",
         docstring="output dimension")
+    if self.layer_class in self._LayerClassesWithExplicitTarget:
+      self.params["target"] = LayerSignature.Param(
+        self,
+        inspect.Parameter(
+          name="target",
+          kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
+        param_type_s=LayerRef,
```
albertz, line 276: This is supposed to be a string.
```suggestion
        param_type_s="LayerRef",
```

models/_generate_layers.py
albertz: Also, `LayerRef` not needed here.
```suggestion

```

models/_generate_layers.py
```diff
@@ -263,6 +267,14 @@ def _init_args(self):
           kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
         param_type_s="int",
         docstring="output dimension")
+    if self.layer_class in self._LayerClassesWithExplicitTarget:
+      self.params["target"] = LayerSignature.Param(
+        self,
+        inspect.Parameter(
+          name="target",
+          kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
```
albertz, line 275: This should be `KEYWORD_ONLY`.
```suggestion
          kind=inspect.Parameter.KEYWORD_ONLY),
```

albertz:
> Not all layers should have the possibility to set targets, since it does not make sense in most cases.

Actually this is wrong. Currently all layers can have losses, and the target usually is used for the loss. E.g. when you set `loss="ce"`, this would define the targets for this loss.

However, my suggestion was to anyway define losses more in PyTorch-style, i.e. as-is (via some `CrossEntropy` module or so), and then we have the function `LayerRef.mark_as_loss`.

This was one of the reasons I excluded `target` for now.

Then, I also was not really sure whether we should have a maybe better way to access extern data. I'm not sure our distinction between extern data (`ExternData` in RETURNN, and the special layers `data:<...>`, and also the targets) and normal layers is really so useful or only complicates things. Esp, I don't like that these are global names for the whole network. I think @JackTemaki also doesn't like this too much.

But then, I also don't really have a better solution. For accessing extern data, I made the special function `get_extern_data`, so that it will be easier to refactor it later maybe, or change the internals. I don't want that anyone needs to write `data:<...>` explicitly at any point.

`targets` now is bad in a similar way. But I see that we need it specifically for `BaseChoiceLayer`. So I guess this PR (with my proposed changes) is still fine. Unless someone has a better suggestion right now.

models/_generate_layers.py
```diff
@@ -253,6 +254,9 @@ def module_call_needs_suppress_shadow_builtin_warning(self):
     LinearLayer, ConvLayer, TransposedConvLayer, RecLayer, RnnCellLayer,
     PositionalEncodingLayer, RelativePositionalEncodingLayer}
 
+  _LayerClassesWithExplicitTarget={
```
albertz, line 257: See failing tests.
`returnn_common/models/_generate_layers.py:257: WEAK WARNING Pep8CodeStyle: E225: missing whitespace around operator`

```suggestion
  _LayerClassesWithExplicitTarget = {
```

Atticus1806:
> However, my suggestion was to anyway define losses more in PyTorch-style, i.e. as-is (via some CrossEntropy module or so), and then we have the function LayerRef.mark_as_loss.

Yes in my statement regarding the target I was already taking the change in the style we will handle the loss in the future into account. Should have clarified that.

Also I just realized that I didn't have the i6 Style enabled on the device I was using, changed that now, sorry for the style bugs.

>Esp, I don't like that these are global names for the whole network.
Yes, I was having a conversation with @JackTemaki about that topic yesterday aswell, since it might also directly affect the way to which function the target is passed. For now we decided that giving it in the `__init__` is the most coherent way with current configs. Once there is a major change in Returnn (maybe with a behaviour version?) this might change again.

models/_generate_layers.py
```diff
@@ -263,6 +266,14 @@ def _init_args(self):
           kind=inspect.Parameter.POSITIONAL_OR_KEYWORD),
         param_type_s="int",
         docstring="output dimension")
+    if self.layer_class in self._LayerClassesWithExplicitTarget:
+      self.params["target"] = LayerSignature.Param(
+        self,
+        inspect.Parameter(
+          name="target",
+          kind=inspect.Parameter.KEYWORD_ONLY),
+        param_type_s="LayerRef",
```
albertz, line 275: I'm just realizing that `LayerRef` is the wrong type here. `str` would be correct.

Although I'm not sure whether it actually maybe should be a `LayerRef`. See my other comment.

albertz:
>  For now we decided that giving it in the `__init__` is the most coherent way with current configs.

I just realized this aspect. I'm not sure this is good. All data (layer refs, and target is just a special layer ref) are referenced explicitly in the `forward` / `make_layer_dict` function. `target` would be somewhat inconsistent now. Maybe `target` should be handled just exactly like other layer refs.

albertz:
Btw, in the future, better do not merge yet until the PR is approved by someone.

Atticus1806:
>Btw, in the future, better do not merge yet until the PR is approved by someone.

Oh I do have rights to "just merge"? My bad. It was not intentional, when I saw the "Merge button" this morning I thought it wasn't there before and appeared because you accepted the PR and since you wrote that the PR is fine for you I didn't question that any further. I'm sorry.

albertz:
> > Btw, in the future, better do not merge yet until the PR is approved by someone.
> 
> Oh I do have rights to "just merge"? My bad. It was not intentional, when I saw the "Merge button" this morning I thought it wasn't there before and appeared because you accepted the PR and since you wrote that the PR is fine for you I didn't question that any further. I'm sorry.

No problem. We can just open a new PR for further changes.

I'm still not 100% sure how it really would be optimal. Maybe like I commented.

Atticus1806:
I think both could be fine. Maybe I can give a bit more context to the thinking process and what I discussed with @JackTemaki yesterday: Putting `target` into the forward would make exceptions necessary, since then it must not be resolved like the other `LayerRef`s, but rather be handled as a raw string. This then means that either there is a real exception for a `LayerRef` called `target` or that target after all is just a string that is passed. If its just a string then it does not make a lot of difference to putting the target into the `init`, since Layers using targets (right now only `BaseChoice`) should only be called once and the string (or `get_special_layer(string)`) of the target would already be defined when calling the `init` just without the necessary exceptions. But maybe there is stuff we oversaw, what do you think?

albertz:
> Putting `target` into the forward would make exceptions necessary, since then it must not be resolved like the other `LayerRef`s, but rather be handled as a raw string.

I don't exactly understand. What exceptions? (`Exceptions`s?) Why are they necessary? Why then as a raw string?

When it becomes an argument of `forward`/`make_layer_dict`, then it should not be a `str`, but a `LayerRef`, and just handled like other layer refs. In the config, you would use `target=get_extern_data("classes")` or so.

When it is an argument of `__init__`, then it should not be a `LayerRef` but a `str`. I think there never should be a `LayerRef` argument to `__init__` actually.

JackTemaki:
> Putting `target` into the forward would make exceptions necessary, since then it must not be resolved like the other `LayerRef`s, but rather be handled as a raw string

I was talking about exceptions in the generation script, not about exceptions for resolving the layers (this is probably already correct, and for external data this does not matter). So the question is if we need to implement the "ChoiceLayer" by hand, or just update the generation code. I did not look into implementation details here yet though...

albertz:
Ah you don't really mean exceptions (`Exception`) here but just whether it needs special handling. But the whole PR is already about this, and we already have this for other things (e.g. `n_out`), so I thought this is clear. Some special handling is fine.

We should also not care too much how it is done internally and also not on RETURNN side. Esp, when something is inconsistent on RETURNN side, we do not need to replicate it exactly in the same way here in this wrapper, but we can make it nicer. I think `target` is one such example. `target` should not be different from other layer refs. Even on RETURNN side, internally, they are mostly equivalent. Just the syntax is slightly different, but the syntax is what we abstract away here anyway.

Atticus1806:
Also I now realize I was mixing things up in my understanding of how some of the building works, so this probably also brought in additional confusion.

So if I understand you correctly we would instead of making changes to the `__init__` make changes to the `make_layer_dict` function. Does that mean we would have to change `Choice` as a Layer (and potentially other layers inheriting from `BaseChoice` or what would be the smarter way to go about this? Or would you just suggesting putting the `Choice` Layer in to a handwritten File?

albertz:
> So if I understand you correctly we would instead of making changes to the `__init__` make changes to the `make_layer_dict` function.

From this generator-script, this is almost the same. See `is_module_init_arg` and `is_module_call_arg`.

> Does that mean we would have to change `Choice` as a Layer (and potentially other layers inheriting from `BaseChoice` or what would be the smarter way to go about this? Or would you just suggesting putting the `Choice` Layer in to a handwritten File?

No, none of that.

The only thing about `BaseChoiceLayer` is that it would enable this explicit `target` argument (as you already had it in this PR). Maybe also `LossLayer` (and maybe `CrossEntropyLayer` etc) would be added later (although not sure yet. Maybe `LossLayer` should be ignored...).

And then we would just have some generic handling for `target`.

I can prepare this.

Issue #9, opened 2021-06-23
Loss Scale


Atticus1806:
So as I understand right now, we want to go away from using `Returnn` Losses and use Loss `'as_is'`. How would adding a loss scale be intended here? Use a mult layer with a constant? Or is there a better way or not implemented yet? Maybe add a parameter to the `mark_as_loss()` function with default value 1?

albertz:
> How would adding a loss scale be intended here?

I have not thought about this yet.

> Use a mult layer with a constant?

Yea, this is one way which is always possible of course. But not sure if the best way. (Although, I think this would most closely match how you would do it in PyTorch.)

> Or is there a better way or not implemented yet?

You know what is implemented. :) But in any case, if sth is not implemented, we can always simply just implement it.

> Maybe add a parameter to the `mark_as_loss()` function with default value 1?

Yea, that is also a good idea. PR is welcome.

Atticus1806:
Just opened a PR.

>You know what is implemented. :)

Yes you are right, I should know or can read it up, sometimes I am not sure wether I overlooked something. But I guess in `returnn_common` since its quite new and not really used yet I can safely assume that if I dont find something it is not there yet. :)

Atticus1806:
I will close this now, we can discuss any changes in the PR itself I guess.

albertz:
> I will close this now, we can discuss any changes in the PR itself I guess.

An issue should only be closed once the issue/bug is really resolved/fixed. E.g. when a corresponding PR is merged.

A PR (or commit) can also automatically close an issue when you write sth like "Fix #9" in its description.

PR #10, opened 2021-06-24
added loss scale

Atticus1806:
Added loss scale parameter to the `mark_as_loss` function in a similiar style to how it already sets `loss = as_is` . Cf. #9

### DIFF
--- a/models/base.py
+++ b/models/base.py
@@ -123,7 +123,6 @@ class Layer(LayerRef):
   def mark_as_loss(self, loss_scale: Optional[float] = 1.0):
     """
     Mark this as a loss.
-    :param loss_scale: scale factor for loss (1.0 by default)
     """
     assert "loss" not in self.layer_dict
     self.layer_dict["loss"] = "as_is"
--- a/models/base.py
+++ b/models/base.py
@@ -120,12 +120,15 @@ class Layer(LayerRef):
     self.maker = maker
     self.layer_dict = layer_dict
 
-  def mark_as_loss(self):
+  def mark_as_loss(self, loss_scale=1.0):
     """
     Mark this as a loss.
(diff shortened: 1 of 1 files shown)


models/base.py
```diff
@@ -120,12 +120,15 @@ def __init__(self, maker: ILayerMaker, layer_dict: LayerDictRaw):
     self.maker = maker
     self.layer_dict = layer_dict
 
-  def mark_as_loss(self):
+  def mark_as_loss(self, loss_scale=1.0):
```
albertz, line 123: Better:
```suggestion
  def mark_as_loss(self, loss_scale: Optional[float] = 1.0):
```

models/base.py
```diff
@@ -120,12 +120,15 @@ def __init__(self, maker: ILayerMaker, layer_dict: LayerDictRaw):
...
     """
     Mark this as a loss.
+    :param loss_scale: scale factor for loss (1.0 by default)
```
albertz, line 126: Make an empty line after the description before the first `:param` (consistency).

Or maybe even better: Just remove this line. It doesn't really add any useful information.
```suggestion
```

models/base.py
```diff
@@ -120,12 +120,15 @@ def __init__(self, maker: ILayerMaker, layer_dict: LayerDictRaw):
...
     """
     assert "loss" not in self.layer_dict
     self.layer_dict["loss"] = "as_is"
+    assert "loss_scale" not in self.layer_dict
+    self.layer_dict["loss_scale"] = loss_scale
```
albertz, line 131:
```suggestion
    if loss_scale is not None:
      assert "loss_scale" not in self.layer_dict
      self.layer_dict["loss_scale"] = loss_scale
```

albertz (approved):

Issue #7, opened 2021-06-15
Accessing layers not returned by the forward function of a module --or-- returning multiple layer-refs


Atticus1806:
Is there currently a way to access layers (or rather the corresponding `LayerRef`) of a Layer used in a Module but not returned by the forward function. I want to keep my Modules as generic as possible, always returning more than one layer there seems too much for me, if its just used in one or two settings.

albertz:
No.

Just like we want to add `LayerRef.__add__`, `__sub__` etc (so that you can write `x + y` directly), we could also add some function `LayerRef.get_sub_layer(name: str) -> LayerRef` or sth like that, to access sub layers (assuming it's inside a subnet).

However, in general, I wonder if this is a good idea to use too often. Because this would then use assumptions on the internal structure of `forward` and on the returned layer, and also the naming logic. This could easily lead to bugs, or break at some point.

Also look at PyTorch (the whole API is heavily inspired from PyTorch and its `Module`). You also cannot access any intermediate outputs in its `forward` function. If there is some sensible output, even if only rarely needed, it would still be part of the output. E.g. look at [LSTM](https://pytorch.org/docs/stable/generated/torch.nn.LSTM.html), it returns `(output, (h_n, c_n))`, and usually you only use `output`, and often write `output, _ = LSTM(...)(x)` or so.

Instead of returning a tuple, you could also return a dict, or namedtuple, which maybe would make it more structured.

Another suggestion: Maybe you could also have an argument `return_extra_x=True` on your `forward` function, and if enabled, it additionally returns whatever you need.

Atticus1806:
I think it makes sense what you say. Changed my code to return multiple layers according to needs.

Atticus1806:
I have to reopen this, because now when starting to debug my first "finished" model I run into problems while returning multiple layers. Maybe I am doing something wrong, so I am trying to show all the steps I am doing:
Lets assume I have a very simple Module: 
```
class EncoderConv(Module):
  def __init__(self, l2=1e-07, n_out=256, dropout=0.5):
    super().__init__()
    self.conv_0 = ConvBlock(l2=l2, n_out=n_out, dropout=dropout)
    self.conv_1 = ConvBlock(l2=l2, n_out=n_out, dropout=dropout)
    self.conv_2 = ConvBlock(l2=l2, n_out=n_out, dropout=dropout)

  def forward(self, embedding: LayerRef) -> LayerRef:
    x_ = self.conv_0(embedding)
[...]

PR #13, opened 2021-06-25
Multiple Returns in Forward

Atticus1806:
First draft for a fix for #7 

Not finished yet, comments on what should be done differently most welcome.


Atticus1806:
It might also be that my approach is quite bad and should be done entirely different, then some help on how to approach it better would be good too.

models/base.py
```diff
@@ -166,6 +166,10 @@ def __call__(self, *args, name: Optional[str] = None, **kwargs) -> Layer:
       if self.calls:
         name_ctx.is_repeated_call = True
       layer_dict = self.make_layer_dict(*args, **kwargs)
+      ret = None
+      if isinstance(layer_dict, Tuple):
```
albertz, line 170: There should not be such an `isinstance(..., Tuple)` check, but this should be done via `nest` instead.

albertz: Ah whoops sorry, no this is fine here but the check should be different.

models/base.py
```diff
@@ -166,6 +166,10 @@ def __call__(self, *args, name: Optional[str] = None, **kwargs) -> Layer:
...
+        ret = layer_dict[1:]
+        layer_dict = layer_dict[0]
```
albertz, line 172:
```suggestion
      if not isinstance(layer_dict, dict):
        ret, layer_dict = layer_dict
        assert isinstance(layer_dict, dict)
```

models/base.py
```diff
@@ -179,6 +183,8 @@ def __call__(self, *args, name: Optional[str] = None, **kwargs) -> Layer:
           layer_dict["reuse_params"] = self.calls[0].get_name()
       layer = Layer(self, layer_dict)
       self.calls.append(layer)
+      if ret:
+        return (layer, *ret)
```
albertz, line 187: The return is wrong. Also, the brackets are not needed (and thus should not be used).
```suggestion
        return layer, ret
```
Also, the type annotation of the function (`__call__`) should also be changed.

models/base.py
```diff
@@ -196,8 +202,16 @@ def make_layer_dict(self, *args, **kwargs) -> LayerDictRaw:
     assert name_ctx.maker is self
     name_ctx.is_subnet_ctx = True
     res = self._subnet_func(*args, **kwargs)
-    Copy()(res, name="output")
-    return self._make_layer_dict_from_subnet_ctx(name_ctx)
+    # check if we receive more than one layer as return of a subnet
+    if isinstance(res, (Layer, type(List[Layer]))):
```
albertz, line 206: I don't understand this check? First you want to check for the case that you get a single output as usual.
Also, you do not want to check for `Layer` but `LayerRef`.
```suggestion
    if isinstance(res, LayerRef):
```

models/base.py
```diff
@@ -196,8 +202,16 @@ def make_layer_dict(self, *args, **kwargs) -> LayerDictRaw:
...
+      Copy()(res, name="output")
+      return self._make_layer_dict_from_subnet_ctx(name_ctx)
+    else:
+      # we return more than one layer (thus also working on other layers of the subnet, that are not output
+      # by convention: first layer is the output layer
+      assert isinstance(res, Tuple)
```
albertz, line 212: Remove the assert.

Also, use `nest` here as I wrote before. (`nest.flatten` and then `nest.pack_sequence_as`. As I said, see other related code, e.g. in pytorch-to-returnn.)

models/base.py
```diff
@@ -196,8 +202,16 @@ def make_layer_dict(self, *args, **kwargs) -> LayerDictRaw:
...
+      Copy()(res[0], name="output")
+      return (self._make_layer_dict_from_subnet_ctx(name_ctx), *res[1:])
```
albertz, line 214: Remove the brackets.
Also, return type is wrong.
```suggestion
      return self._make_layer_dict_from_subnet_ctx(name_ctx), nest.pack_sequence_as(...)
```
Also, you need to fix the return type annotation of the function (`make_layer_dict`).

models/base.py
albertz:
```suggestion
     res_flat = nest.flatten(res)
```

models/base.py
```diff
@@ -196,8 +202,16 @@ def make_layer_dict(self, *args, **kwargs) -> LayerDictRaw:
...
-    return self._make_layer_dict_from_subnet_ctx(name_ctx)
+    # check if we receive more than one layer as return of a subnet
+    if isinstance(res, (Layer, type(List[Layer]))):
+      Copy()(res, name="output")
+      return self._make_layer_dict_from_subnet_ctx(name_ctx)
+    else:
+      # we return more than one layer (thus also working on other layers of the subnet, that are not output
+      # by convention: first layer is the output layer
+      assert isinstance(res, Tuple)
+      Copy()(res[0], name="output")
```
albertz, line 213:
```suggestion
      Copy()(res_flat[0], name="output")
```

albertz:
Can you add a simple test case? (Just `def forward(self, x): return x, x` as the most simple. And maybe slightly more complex if you think that simple one doesn't cover all cases. But doesn't need to be an overly complex nested structure. We can trust that `nest` does the right thing here.)

models/base.py
Atticus1806: I am not sure if `layer_dict, ret = layer_dict` works in the general case, where possibly there is more than just one additional layer passed outside of the subnetwork, shouldn't this run into an error in a case where `len(layer_dict)==3` ?

models/base.py
Atticus1806: Yes I realize the typing is wrong, I did not think about that yet, but I feel like this is something we can fix once the code is "done"
The brackets where added for one of the test i think, since they test in python 3.7 iirc, and from what I understand, returning without brackets is only possible since 3.8. But maybe I am wrong on this part. 
If we only do `return ret` then we dont return the layer that is intended to be the output of the subnetwork, since `ret = layer_dict[1:]` or am I missing something?

models/base.py
Atticus1806: You are right about the `LayerRef` I think. I am checking for `List` since the input into `from` for `returnn` might also be a list of layer (which by default then are concated). Maybe I am missing a point here but just checking for `LayerRef` will miss these cases, right?

models/base.py
Atticus1806: I will fix the annotation tomorrow but I agree on the suggested changes, so I will commit them and add the typing tomorrow.

Atticus1806:
>Can you add a simple test case? 

Yes I can do that tomorrow, I agree, just adding a simple case with your suggestion should do the job, after reading through into `nest` I agree we can rely on that.

models/base.py
Atticus1806: Now that I think about it, I am not quite sure again. `nest.pack_sequence_as` from my understanding restores the structure of the flattened sequence but doesn't that mean that in this case we return the output layer twice, since it is the result of a Copy layer of the first entry of `res`. Is this intended or am I not getting something here? My current understand would want to do something like `nest.pack_sequence_as(res[1:], res_flat[1:])`.

Atticus1806:
This should be what you  thought of for the test right? I am not sure about the assert at the end if that is even needed, edcided to keep it in for now.

models/base.py
albertz: `ret` itself would be the tuple, or whatever other structure.

models/base.py
albertz: > Yes I realize the typing is wrong, I did not think about that yet, but I feel like this is something we can fix once the code is "done"

Better do it directly now that we already have it here and do not forget about it.

> The brackets where added for one of the test i think, since they test in python 3.7 iirc, and from what I understand, returning without brackets is only possible since 3.8. But maybe I am wrong on this part.

Returning without brackets was possible since Python 1 I think.

> If we only do `return ret` then we dont return the layer that is intended to be the output of the subnetwork, since `ret = layer_dict[1:]` or am I missing something?

Yes exactly. We instead would return the same structure which is intended to be returned, exactly what `forward` returned, i.e. a tuple of layers, or whatever structure `forward` returned.

models/base.py
albertz: This is the return, not the input.

models/base.py
albertz: No, we return the layer dict + the nested structure (of layer refs etc).

tests/test_models_layers.py
```diff
@@ -187,3 +187,22 @@ def forward(self, x: LayerRef) -> LayerRef:
 
   assert "TestBlock" in net_dict and "TestBlock_0" in net_dict
   assert_equal(net_dict["TestBlock_0"]["reuse_params"], "TestBlock")
+
+
+def test_mult_returns():
```
albertz, line 192: `mult` usually means "multiplication". Better write `multiple`.

models/base.py
Atticus1806: Ah I see. 👍

tests/test_models_layers.py
```diff
@@ -187,3 +187,22 @@ def forward(self, x: LayerRef) -> LayerRef:
...
+  class _Net(Module):
+    def __init__(self):
+      super().__init__()
+      self.linear = Linear(n_out=42, activation=None)
+
+    def forward(self) -> Union[LayerRef, Tuple[LayerRef, Any]]:
```
albertz, line 198:
```suggestion
    def forward(self) -> Tuple[LayerRef, LayerRef]:
```

tests/test_models_layers.py
```diff
@@ -187,3 +187,22 @@ def forward(self, x: LayerRef) -> LayerRef:
...
+      """
+      Forward
+      """
+      x = get_extern_data("data")
+      x = self.linear(x)
+      return x, x
```
albertz, line 204: The main network should not return multiple values. This should be some other module.

tests/test_models_layers.py
```diff
@@ -187,3 +187,22 @@ def forward(self, x: LayerRef) -> LayerRef:
...
+  class _Net(Module):
+    def __init__(self):
+      super().__init__()
+      self.linear = Linear(n_out=42, activation=None)
+
+    def forward(self) -> Union[LayerRef, Tuple[LayerRef, Any]]:
+      """
+      Forward
+      """
+      x = get_extern_data("data")
```
albertz, line 202: Better pass `x: LayerRef` as an argument to `forward`, to make this more like a standard module.

models/base.py
Atticus1806: >Returning without brackets was possible since Python 1 I think.

Hmm there was an error with one commit without these brackets, I think it was related to syntax including * so e.g. `layer, *ret` I think that failed in one of the tests, but we removed that anyways

>Yes exactly. We instead would return the same structure which is intended to be returned, exactly what forward returned, i.e. a tuple of layers, or whatever structure forward returned.

Ahh I think I understand: So because in the `make_layer_dict` function we would return as the second argument all of the layers passed, this would here then mean that the output is also in there so we would not lose any reference to any layer right? So if `forward` returned `out, (hidden, last)` then `make_layer_dict` returns `LayerDict, (out, (hidden, last))` right? Then it makes sense to me.

models/base.py
Atticus1806: Oh you are right, whoops.

models/base.py
Atticus1806: I think I understand it now, cf. other comments.

Atticus1806:
Regarding the test: would you even assert something like this? Or would you check for something else? 

>The main network should not return multiple values. This should be some other module.
>Better pass x: LayerRef as an argument to forward, to make this more like a standard module.
Do you prefer it like the way I pushed it now?

models/base.py
albertz: > Hmm there was an error with one commit without these brackets, I think it was related to syntax including * so e.g. `layer, *ret` I think that failed in one of the tests

Ah yes, that's possible.

> Ahh I think I understand: So because in the `make_layer_dict` function we would return as the second argument all of the layers passed, this would here then mean that the output is also in there so we would not lose any reference to any layer right?

Yes. In the first case with only one layer, this becomes the "output" of the subnetwork, and then the subnetwork itself is fine as a layer-ref (which doesn't need to be explicitly specified because it gets the layer dict).

In the case with multiple outputs / some nested structure, we just pass on all layer refs as-is. The "output" of the subnetwork, and the subnetwork layer itself would actually not directly be used but just the layer refs directly (e.g. like "subnet/out", "subnet/hidden", "subnet/last", whatever).

> So if `forward` returned `out, (hidden, last)` then `make_layer_dict` returns `LayerDict, (out, (hidden, last))` right? Then it makes sense to me.

Yes exactly. And this `ILayerMaker.__call__` would return again only the `out, (hidden, last)`.

tests/test_models_layers.py
```diff
@@ -187,3 +187,34 @@ def forward(self, x: LayerRef) -> LayerRef:
 
   assert "TestBlock" in net_dict and "TestBlock_0" in net_dict
   assert_equal(net_dict["TestBlock_0"]["reuse_params"], "TestBlock")
+
+
+def test_multiple_returns():
+  class _SubNet(Module):
+    def __init__(self):
+      super().__init__()
+      self.linear = Linear(n_out=42, activation=None)
```
albertz, line 196: I would use smaller dimensions for a test (n_out < 20 or so).
Also, this is already slightly more complex than I suggested. I suggested it even without this `linear`. But anyway, this is also fine.

tests/test_models_layers.py
```diff
@@ -187,3 +187,34 @@ def forward(self, x: LayerRef) -> LayerRef:
...
+      super().__init__()
+      self.sub = _SubNet()
+
+    def forward(self) -> LayerRef:
+      """
+      Forward
+      """
+      x = get_extern_data("data")
+      out, add_out = self.sub(x)
+      assert out == add_out[0]
```
albertz, line 216: This is wrong.
```suggestion
      assert out == add_out
```

tests/test_models_layers.py
```diff
@@ -187,3 +187,34 @@ def forward(self, x: LayerRef) -> LayerRef:
...
+      return out
+
+  net = _Net()
+  net.make_root_net_dict()
```
albertz, line 220: Also store the dict, and print it, verify it by hand (by looking at it), and then add some sanity checks.
```suggestion
  net_dict = net.make_root_net_dict()
  pprint(net_dict)
```

tests/test_models_layers.py
albertz: Actually, I think removing this `linear` might trigger some strange other case. So leave it in (but maybe change dim).
```suggestion
      self.linear = Linear(n_out=13, activation=None)
```

albertz:
Note: Now we have this failing:
```
  File "/home/runner/work/returnn_common/returnn_common/models/base.py", line 446, in NameCtx.get_name_in_current_ctx
    line: assert len(cur_scope_abs) >= len(self_name_abs)  # not implemented otherwise
    locals:
      len = <builtin> <built-in function len>
      cur_scope_abs = <local> [<NameCtx maker:<_Net> name:None root:7f17e0882ed0>]
      self_name_abs = <local> [<NameCtx maker:<_Net> name:None root:7f17e0882ed0>, <NameCtx maker:<_SubNet> name:None/'sub' root:7f17e0882ed0>, <NameCtx maker:<Linear> name:None/'sub'/'linear' root:7f17e0882ed0>]
AssertionError
```
This is because it's just not implemented otherwise.
The error might look a bit cryptic. But this is just the logic to get the right name for the layer ref in the current context. `os.path.relpath` would be the analoge for file paths. Maybe we could even reuse the logic of that function (and replacing `../` by `base:`).

The list here corresponds to the elements of the absolute name path, where each element is a `NameCtx` (which is like a directory). The first entry with `name:None` is the root entry (which basically corresponds to `/`).

The current implementation basically just covers the case of adding `base:` to the name, but not going down again.
I also feel that the current implementation is maybe implemented somewhat too complicated.

Atticus1806:
I am trying to understand the function right now. So the case where it currently not works is when the layer that we want to access is actually in a layer "deeper" than the layer we are currently in right? So what is missing is a routine that resolves this by building the path down to the sublayer?

albertz:
> I am trying to understand the function right now. So the case where it currently not works is when the layer that we want to access is actually in a layer "deeper" than the layer we are currently in right? So what is missing is a routine that resolves this by building the path down to the sublayer?

Yes. The function `get_name_in_current_ctx` just needs to be extended for that case.

Atticus1806:
I just pushed a try of this. The test case works, but I think there is one more case that needs to be handled: Right now I think is keeps the absolute name from the root, so maybe we need to "shorten" the beginning aswell but on the spot I am not sure how to find the "length" to the root.

models/base.py
```diff
@@ -430,7 +443,12 @@ def get_name_in_current_ctx(self) -> str:
     cur_scope_abs = cur_scope.get_abs_name_ctx_list()
     self_name_abs = self.get_abs_name_ctx_list()
     assert cur_scope_abs[0] is self_name_abs[0]  # same root
-    assert len(cur_scope_abs) >= len(self_name_abs)  # not implemented otherwise
+    if len(cur_scope_abs) < len(self_name_abs): # grabbing a sublayer
```
albertz, line 446: Better do it the way as I suggested.
I.e. first:
`max_common_len = max(len(cur_scope_abs), len(cur_scope_abs))`
Then for the `while` loop below, add also `common_len < max_common_len and ` in the beginning.
And then add your code below for the case `common_len < len(self_name_abs)`.

Atticus1806: Ahh just commited another change before I saw your comment :D, but sure we can do it like this, let me restructe the code.

Atticus1806: This crashes with index out of range though. Should it be `common_len+1 < max_common_len` ? Or we need to change the second part of the while statement.

Atticus1806:
Also I made a small update to the testcase adding not only a depth 1 dependency but also a dependency of depth 2, since some of my intial versions worked for depth one but not for 2 (assuming 2 is enough, if depth 2 works 3 and so on should also)

models/base.py
albertz: It's very possible that I have made some +- 1 or 2 error here. :) It's a bit late and I'm going to sleep now.

Atticus1806: Same :D, lets talk about this tomorrow/Monday

albertz:
> Also I made a small update to the testcase adding not only a depth 1 dependency but also a dependency of depth 2, since some of my intial versions worked for depth one but not for 2 (assuming 2 is enough, if depth 2 works 3 and so on should also)

Better make two separate test cases then. One for the most straightforward case, and then for your new case.

Atticus1806:
Okay nvm. I made a mistake. ofc `data:data` does not need `base:` infront of it 😅  not clue what I was thinking there will fix the tests and the code generation i guess

Atticus1806:
Wait now that I see, it is also checked in another test. Does that mean we are changing something here from returnn? because there the `base:` prefix is not needed right?

tests/test_models_layers.py
```diff
@@ -219,7 +219,7 @@ def forward(self) -> LayerRef:
   net_dict = net.make_root_net_dict()
   pprint(net_dict)
   assert net_dict["output"]["from"] == "sub/linear"
-  assert net_dict["sub"]["subnetwork"]["linear"]["from"] == "base:data:data"
+  assert net_dict["sub"]["subnetwork"]["linear"]["from"] == "data:data"
```
albertz, line 222: But shouldn't it be `"base:data:data"`? It should get the extern data from the root net, not from the subnet.

albertz: > Wait now that I see, it is also checked in another test. Does that mean we are changing something here from returnn? because there the `base:` prefix is not needed right?

It is needed. Why should it not be needed? You want to get `data` from the root net.

Atticus1806: You are absolutely right, I think I am too tired right now. I forgot about the possible `from:` which we do not use but which is a theoretical possibility replacing data:data. My bad.

Atticus1806:
Okay then this I think is what we want. Just have a look and if there is anything still improveable just comment and I will deal with it tomorrow.

models/base.py
```diff
@@ -430,10 +443,16 @@ def get_name_in_current_ctx(self) -> str:
     cur_scope_abs = cur_scope.get_abs_name_ctx_list()
     self_name_abs = self.get_abs_name_ctx_list()
     assert cur_scope_abs[0] is self_name_abs[0]  # same root
-    assert len(cur_scope_abs) >= len(self_name_abs)  # not implemented otherwise
     common_len = 0
-    while cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
+    max_common_len = max(len(cur_scope_abs), len(cur_scope_abs))
+    while common_len+1 < max_common_len and cur_scope_abs[common_len+1] is self_name_abs[common_len+1]:
```
albertz, line 448: Keep code style consistent.
```suggestion
    while common_len + 1 < max_common_len and cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
```

models/base.py
```diff
@@ -430,10 +443,16 @@ def get_name_in_current_ctx(self) -> str:
...
       common_len += 1
+    if common_len < len(self_name_abs) and len(cur_scope_abs) < len(self_name_abs):  # grabbing a sublayer
```
albertz, line 450: This should not be a separate case. Then you would still not cover all cases (e.g. first going up and then going down). Just merge it with the code below. I think it would be very similar to the `relpath` code which I posted.

models/base.py
```diff
@@ -424,18 +437,27 @@ def get_name_in_current_ctx(self) -> str:
     """
     Get layer name valid for current scope.
     """
+    """
+    Get layer name valid for current scope.
+  """
```
albertz, line 442:
```suggestion
```
Copy & paste error?

models/base.py
```diff
@@ -424,18 +437,27 @@ def get_name_in_current_ctx(self) -> str:
...
     assert cur_scope_abs[0] is self_name_abs[0]  # same root
-    assert len(cur_scope_abs) >= len(self_name_abs)  # not implemented otherwise
     common_len = 0
-    while cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
+    max_common_len = min(len(cur_scope_abs), len(self_name_abs))
+    while common_len + 1 < max_common_len and cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
       common_len += 1
-    assert common_len == len(self_name_abs) - 2  # not implemented otherwise
-    return "base:" * (len(cur_scope_abs) - len(self_name_abs) + 1) + self.name
+    # assert common_len == len(self_name_abs) - 2  # not implemented otherwise
```
albertz, line 453: Remove code, do not comment it out.
```suggestion
```

models/base.py
```diff
@@ -424,18 +437,27 @@ def get_name_in_current_ctx(self) -> str:
...
+    full_name = self_name_abs[-1].get_abs_name()
+    subs = full_name.split("/")
+    relative_name = subs[len(cur_scope_abs)-1:-1]
+    prefix = "/".join(relative_name)
+    if not prefix == "":
```
albertz, line 458:
```suggestion
    if prefix != "":
```
Or:
```suggestion
    if prefix:
```

models/base.py
```diff
@@ -424,18 +437,27 @@ def get_name_in_current_ctx(self) -> str:
...
+      prefix += "/"
+    return "base:" * (len(cur_scope_abs) - len(self_name_abs) + 1) + prefix + self.name
```
albertz, line 460: I'm pretty sure you miss `common_len` here. See the `relpath` logic. This would correspond to `commonprefix`.

models/base.py
```diff
@@ -430,12 +443,17 @@ def get_name_in_current_ctx(self) -> str:
...
-    assert len(cur_scope_abs) >= len(self_name_abs)  # not implemented otherwise
     common_len = 0
-    while cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
+    max_common_len = min(len(cur_scope_abs), len(self_name_abs))
+    while common_len + 1 < max_common_len and cur_scope_abs[common_len + 1] is self_name_abs[common_len + 1]:
       common_len += 1
-    assert common_len == len(self_name_abs) - 2  # not implemented otherwise
-    return "base:" * (len(cur_scope_abs) - len(self_name_abs) + 1) + self.name
+    full_name = self_name_abs[-1].get_abs_name()
+    subs = full_name.split("/")
```
albertz, line 451: Why so complicated? I think we should not do logic based on the strings (`split("/")`) when we already have it splitted beforehand (`self_name_abs`).

models/base.py
```diff
@@ -430,12 +443,17 @@ def get_name_in_current_ctx(self) -> str:
...
+    relative_name = subs[common_len:-1]
+    prefix = "/".join(relative_name)
+    if prefix:
+      prefix += "/"
+    return "base:" * (len(cur_scope_abs) - len(self_name_abs) + 1) + prefix + self.name
```
albertz, line 457: I also think `common_len` must be used for the calculation of number of `base:`. Again, see the code of `relpath`.

I think sth like this should work (without testing it):

```suggestion
    prefix = "base:" * (len(cur_scope_abs) - common_len + 1)
    postfix = "/".join([ctx.name for ctx in self_name_abs[common_len:]])
    return prefix + postfix

```

albertz:
As we keep iterating over `get_name_in_current_ctx`, I think we should also add some test cases specifically for that function only. This test should test for the 4 relevant cases:

* Same context.
* Going up (adding `base:`).
* Going down (adding `sub/` or so).
* Both going up and down (e.g. adding `base:sub/`).

Atticus1806:
I will create a test for that, consisting of a network which uses all of these 4 variants.

albertz:
No, not an indirect test via network. A direct test for the function.


Benedikt Hilmes ***@***.***> schrieb am Do., 1. Juli 2021,
09:53:

albertz:
But you can still leave the other tests as well.

Albert Zeyer ***@***.***> schrieb am Do., 1. Juli 2021, 11:46:

albertz:
Btw, when you think this PR is ready, then mark it as ready (remove the draft status).

albertz:
> As we keep iterating over `get_name_in_current_ctx`, I think we should also add some test cases **specifically for that function only** (without creating any network, `Module`, anything, just calling this function only, nothing else). This test should test for the 4 relevant cases: ...

This is still missing.

Atticus1806:
> Btw, when you think this PR is ready, then mark it as ready (remove the draft status).

Good point. I think the topic why the PR was opened is solved, I just was not sure about adding the tests before removing the draft. I think we should add them before, since otherwise the "task" to make tests might get lost. 

> This is still missing.

Yes you are right, I am kind of struggling with that if I am being honest, trying differnt things (like the new tests I pushed) but I didnt manage to find a way to properly build a test without any Network etc. yet, so I am currently working on that.

albertz:
> I just was not sure about adding the tests before removing the draft. I think we should add them before

Yes right. Add them first.

> > [tests] still missing.
> 
> Yes you are right, I am kind of struggling with that if I am being honest, trying differnt things (like the new tests I pushed) but I didnt manage to find a way to properly build a test without any Network etc. yet, so I am currently working on that.

I think you just need to create a hierarchy of `NameCtx` objects which correspond to the cases. Maybe you could create some helper functions (only for the tests, also only part of the test suite) for that. This shouldn't be difficult.

Atticus1806:
>I think you just need to create a hierarchy of NameCtx objects which correspond to the cases. Maybe you could create some helper functions (only for the tests, also only part of the test suite) for that. This shouldn't be difficult.

Yes after all you were right. It took me a while to figure out how to build that hierarchy, but now I have a first draft. I pushed it here so you can comment on what you would improve, since I am sure that it is not clean yet. As soon as we finished reviewing that test I think we might be good to go with this PR.

albertz (approved):

albertz:
Looked good now. :) Thanks for the work.

Issue #11, opened 2021-06-25
Beam Size in Choice and BaseChoice


Atticus1806:
Both `BaseChoice` and `Choice` have the parameter `beam_size`, which causes the following error when building a `Choice` layer:
`TypeError: __init__() missing 1 required keyword-only argument: 'beam_size'`

From reading the code, I would suggest removing beam_size from the `Choice` layer (class), since other classes inheriting from `BaseChoice` might also need it. What do you think?

Also how would you handle this? Make an exception in the `_generate_layers` script or is there a better way? 
I can make a PR once we decided on a good solution.

albertz:
This could be similarly handled as I suggested in #12. I.e. detect this more automatically. We would need to find and parse the super(...).__init__(...) call in the code, and there we could see that `beam_size=beam_size`, and we would replicate that in the generated code.

albertz:
Fixed via #14.

PR #14, opened 2021-07-15
Parse super call of a layer

### LINKED ISSUE #12: Kind parameter in Eval Layer
Currently the `kind` parameter for the Eval layer needs to be set to `"eval"` by hand, otherwise the `super()` call to the `Combine` layer is missing a parameter. Maybe it would be a good idea to add this as a default parameter to the `Eval` layer? This should not change in any usecase of the eval layer.

Again: Does this require some special case handling (if so, again through modifying `_init_args`) or can we handle it in another way?

Atticus1806:
Trying to fix #12 and #11 
`batch_norm` is excluded manually, since I did not want to modify `_IgnoreParamNames` but caused trouble for `BatchNorm` layer.


models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  call_pruned = call[9:-1]
+
+  # remove excluded params
+  split = call_pruned.split(",")
+  tup_ls = list(map(lambda x: x.split("="), split))
+  for idx, param in enumerate(tup_ls):
+    if param[0].strip() in LayerSignature.IgnoreParamNames or param[0].strip() == "batch_norm":
+      tup_ls.pop(idx)
+  tup_ls = list(map(lambda x: "=".join(x), tup_ls))
+  tup_ls = list(map(lambda x: x.strip(), tup_ls))
```
albertz, line 527: Better use list comprehensions instead.

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  # reformat the super call to extract what we need
+  super_call = lines[0]
+  call = super_call.split(".")[1]
+  call_pruned = call[9:-1]
+
+  # remove excluded params
+  split = call_pruned.split(",")
+  tup_ls = list(map(lambda x: x.split("="), split))
+  for idx, param in enumerate(tup_ls):
+    if param[0].strip() in LayerSignature.IgnoreParamNames or param[0].strip() == "batch_norm":
```
albertz, line 524: `batch_norm` should not be a hardcoded special case here.

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
   return ls
 
 
+def get_super_call_params(layer):
```
albertz, line 500: Define the types (layer and return).

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  # get code as string list
+  code = inspect.getsource(layer.__init__).splitlines()
+
+  # get list of lines starting with super and take the first one
+  lines = [line.strip() for line in code if line.strip().startswith("super")]
+  if not lines:
+    return ''
+
+  # reformat the super call to extract what we need
+  super_call = lines[0]
```
albertz, line 516: It would really help to understand this code if you would add some comment with an example what `super_call` might be now.

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  call = super_call.split(".")[1]
```
albertz, line 517: This is not clear. Again, add a short comment saying what `call` might be now.
```suggestion
  call = super_call.split(".")[1]  # example: call = "..."
```

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  call_pruned = call[9:-1]
```
albertz, line 518: Never use such hardcoded values like `9`. What is 9?

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+
+  # remove excluded params
+  split = call_pruned.split(",")
+  tup_ls = list(map(lambda x: x.split("="), split))
```
albertz, line 522: Use list comprehensions instead.

models/_generate_layers.py
albertz: Also, I assume that you have certain assumptions what the first 9 chars, and the last 1 char of `call` is, right? If you have such assumptions, then always also `assert` for them.

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  for idx, param in enumerate(tup_ls):
```
albertz, line 523: You splitted the `param` before by `=`. Then use it more explicitly here, which makes it easier. E.g.:
```suggestion
  for idx, (key, value) in enumerate(tup_ls):
```

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+    if param[0].strip() in LayerSignature.IgnoreParamNames or param[0].strip() == "batch_norm":
+      tup_ls.pop(idx)
```
albertz, line 525: You cannot do this while iterating over it. It invalidates the order.

albertz: Instead of this loop, simply use a list comprehension.

albertz: Like:
```
tup_ls = [(key, value) for (key, value) in tup_ls if key not in in LayerSignature.IgnoreParamNames]
```

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
...
+  tup_ls = list(map(lambda x: "=".join(x), tup_ls))
+  tup_ls = list(map(lambda x: x.strip(), tup_ls))
+  params = ", ".join(tup_ls)
+  return params
```
albertz, line 529: I don't exactly understand the return. It would return e.g. sth like `"key='value'"`.
This would be used in your code above like `super().__init__({key='value'})`.
But this would be a syntax error. You either would want `super().__init__(dict(key='value'))` or `super().__init__({'key': 'value'})`.

But the return value is anyway not really clear (or clearly documented).

models/_generate_layers.py
```diff
@@ -496,6 +497,38 @@ def collect_layers():
   return ls
 
 
+def get_super_call_params(layer):
+  """
+  Inspects the given layer, extracts unwanted parameters and then returns a string that is ready to be used in the super
+  call.
+  :param layer: Layer to be inspected
+  :return: string with super call parameters
```
albertz, line 505: This is way too vague. And I think you got it wrong. (See comment below.)

albertz (changes requested):
Sorry for the late response.

There are a couple of problems (see individual comments).

Have you actually tried this? I.e. generated the layers code? I wonder if this has worked. Can you also include this in the PR here?

models/_generate_layers.py
Atticus1806: 9 is the length of `"__init__(" and 1 is the length of ")", in the next update of commits I will push this will be clarified through comments and also I replaced the two hardcoded values by `len("__init__(")` and `len(")") in the statement, which makes the code more understandable, but a bit more messy I guess. 
Would be good to hear what you think what we should put here.

models/_generate_layers.py
Atticus1806: Then I am not sure how to handle this case. It is not in `IgnoreParamNames`, but removing it causes an error in `_generated_layers` for `BatchNorm`, because it is not defined there. Or do you suggest adding an additional list with "excluded parameters" which we then can modify/extend instead of just checking batchnorm? 
I guess I will add that, but if its about the "batch_norm" in general I am not sure how to handle.

models/_generate_layers.py
Atticus1806: I think you are missunderstanding something here (which most likely was due to the code being unclear). 
The function returns a pure string, not a dict of some kind, which the directly can be used in the write function. 
I hope this is more clear now.

Atticus1806:
So I just pushed a bunch of changes to the PR, resolving all your comments (hopyfully I didn't miss any).
From what I can see the generation should work, at least for the initial tests I did. I wasn't able to fully test it on a network yet, since there are 1 or 2 other issues with generating the root dictionary, for which I will make a PR/Issue once I found the time to figure out what is going wrong there. But from what I can see this should not be caused by this code, but rather something else. 

I also added a bunch of documentation in the commits, let me know if that makes things more clear.

models/_generate_layers.py
albertz: This was already clear. Maybe you misunderstood what I wrote? It would lead to the code I wrote in my comment (because of `print(f" super().__init__({params})", file=f)` above).

albertz: Ah sorry, I read the code wrongly. The `{`,`}` will not be printed but are used for the f-string. So it would not produce `super().__init__({key='value'})` but `super().__init__(key='value')`. So this looks fine then.

Atticus1806: Most likely I am missinderstanding something 😅 The curly brackets here are just for the format string, I am not sure why that would cause problems.

models/_generate_layers.py
```diff
@@ -81,7 +81,8 @@ def setup():
         if param.is_module_init_arg():
           print(param.get_module_param_docstring(indent="    "), file=f)
       print('    """', file=f)
-      print(f"    super().__init__({'**kwargs' if issubclass(cls_base, LayerBase) else ''})", file=f)
+      params = get_super_call_params(layer_class)
```
albertz, line 84: `params` is a bad variable name (ambiguous or misleading - you would expect a dict, not a str or sth else). Use `params_str` or better `kwargs_repr` or so.

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
   return ls
 
 
+def get_super_call_params(layer: LayerBase) -> str:
```
albertz, line 500: Consistently to the comment above, this function should be called sth like `get_super_call_kwargs_repr` or so.

models/_generate_layers.py
albertz: This is still unclear. What is a "string with super call parameters"?

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+  call = super_call.split(".")[1]  # call = "__init__(beam_size=beam_size, search=search, **kwargs)"
+  assert call.startswith("__init__")
+  assert call.endswith(")")
+  # remove "__init__(" and ") from the call string
+  call_pruned = call[len("__init__("):-len(")")]  # call pruned = "beam_size=beam_size, search=search, **kwargs"
+
+  # get list of tuples for parameter with (param_name, value)
+  tup_ls = [x.split("=") for x in call_pruned.split(",")]
+
+  # ignore **kwargs when excluding
```
albertz, line 528: Again, do not hard code some assumptions without checking them.

But also, this could have been easier directly in the code above:
```
tup_ls = [x.split("=") for x in call_pruned.split(",") if x != "**kwargs"]
```

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+  kwa = tup_ls[-1][0]
+  tup_ls = tup_ls[:-1]
+
+  # additionally excluded parameters
+  excl_params = ["batch_norm"]
+  # remove excluded params
+  tup_ls = ["=".join((key, value)).strip() for (key, value) in tup_ls
+            if (key.strip() not in LayerSignature.IgnoreParamNames and key not in excl_params)]
```
albertz, line 536: Do not indent code like this (hanging indent).

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+
+  # get list of tuples for parameter with (param_name, value)
+  tup_ls = [x.split("=") for x in call_pruned.split(",")]
+
+  # ignore **kwargs when excluding
+  kwa = tup_ls[-1][0]
+  tup_ls = tup_ls[:-1]
+
+  # additionally excluded parameters
+  excl_params = ["batch_norm"]
```
albertz, line 533: I don't understand. Why?

Also, do not hardcode this here.

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+  # remove excluded params
+  tup_ls = ["=".join((key, value)).strip() for (key, value) in tup_ls
+            if (key.strip() not in LayerSignature.IgnoreParamNames and key not in excl_params)]
+  # reattach kwa
+  tup_ls.append(kwa.strip())
```
albertz, line 538: Do not make it so unclear. You expect a certain `kwa`. So do that directly.
```suggestion
  tup_ls += ["**kwargs"]
```

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+  tup_ls = [x.split("=") for x in call_pruned.split(",")]
+
+  # ignore **kwargs when excluding
+  kwa = tup_ls[-1][0]
+  tup_ls = tup_ls[:-1]
+
+  # additionally excluded parameters
+  excl_params = ["batch_norm"]
+  # remove excluded params
+  tup_ls = ["=".join((key, value)).strip() for (key, value) in tup_ls
```
albertz, line 535: Instead of `"=".join((key, value)).strip()`, I think `f"{key}={value.rstrip()}"` is much easier to read.

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+            if (key.strip() not in LayerSignature.IgnoreParamNames and key not in excl_params)]
+  # reattach kwa
+  tup_ls.append(kwa.strip())
+  # join parameters to one string
```
albertz, line 539: I would really avoid comments which do not add any value and just repeat the code. This bloats up the code and thus make it even harder to read in the end.

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
...
+  params = ", ".join(tup_ls)
```
albertz, line 540: Instead of introducing this variable (which is not needed, and introduces one extra line of code, which makes it slightly harder to read), you can also directly return it.

models/_generate_layers.py
```diff
@@ -496,6 +497,50 @@ def collect_layers():
   return ls
 
 
+def get_super_call_params(layer: LayerBase) -> str:
+  """
+  Inspects the given layer in RETURNN, extracts the super call from there, removes unwanted parameters
+  and then returns a string that is ready to be used in the super call for the _generated_layers file.
```
albertz, line 503: What is being returned should be documented under `:return:` and not in the generic description.

There should be an empty line after the generic description.

models/_generate_layers.py
albertz: I don't really know which error you mean. Please be more specific.

albertz: But in any case, this must not be a hardcoded extra case here. This should be more generic.

models/_generate_layers.py
albertz: Code must be understandable. This is a hard requirement.
Also, understandable code is it less messy, not more messy.

models/_generate_layers.py
Atticus1806: In returnn_common we pass the parameters for Batch normalization explicitly via `get_opts` instead of giving them over (as in `RETURNN`) in the parameter `batch_norm`. This used to work, since we were not actually looking at the `super()` withing `RETURNN` which we are doing now.
What is not possible (without other changes I think) is adding `batch_norm` to excluded parameters, since that would interfere with the `_Base` layer.
Maybe I am overlooking something here, but right now I dont see a more generic way, but introducing a new list

models/_generate_layers.py
Atticus1806: Updated the documentation, is it more clear now?

models/_generate_layers.py
```diff
@@ -81,7 +81,8 @@ def setup():
         if param.is_module_init_arg():
           print(param.get_module_param_docstring(indent="    "), file=f)
       print('    """', file=f)
-      print(f"    super().__init__({'**kwargs' if issubclass(cls_base, LayerBase) else ''})", file=f)
+      params = get_super_call_kwargs_repr(layer_class)
```
albertz, line 84: The `params` variable should be renamed (see my earlier comment on this).

models/_generate_layers.py
```diff
@@ -496,6 +497,46 @@ def collect_layers():
   return ls
 
 
+def get_super_call_kwargs_repr(layer: LayerBase) -> str:
+  """
+  Inspects the given layer in RETURNN, extracts the super call from there, removes unwanted parameters.
+
+  :param layer: Layer to be inspected
+  :return: A string containing all arguments of the super() call of the layer which are not excluded.
+           This then can directly be used when writing the super call for the layer for the _generated_layers file
```
albertz, line 506: This is still not clear.
I would write sth like:
```suggestion
  :return: kwargs_repr
    Python source code of kwargs for the super() call of the layer which are not excluded.
    Such that `f"super().__init__({kwargs_repr})"` is valid code.
```

models/_generate_layers.py
```diff
@@ -496,6 +497,46 @@ def collect_layers():
...
+  call_pruned = call[len("__init__("):-len(")")]  # call pruned = "beam_size=beam_size, search=search, **kwargs"
+
+  # get list of tuples for parameter with (param_name, value)
+  tup_ls = [x.split("=") for x in call_pruned.split(",") if x.strip() != "**kwargs"]
+
+  # additionally excluded parameters
+  excl_params = ["batch_norm"]
+  # remove excluded params
+  tup_ls = [
+    f"{key.strip()}={value.strip()}"
```
albertz, line 533: Why do you need `strip` for `key`?

Why do you need `strip` for `value`?

In general, I don't like the `strip` too much at this point anyway. Esp it is bad because you do it three times for `key`, which is repeated code.
I think I would do that separately, above:
```
tup_ls = [(key.strip(), value.strip()) for (key, value) in tup_ls]
```

albertz:
You still miss the generated layers file here in the PR. (Which would be helpful to see what it actually produces.)

models/_generate_layers.py
```diff
@@ -496,6 +497,46 @@ def collect_layers():
   return ls
 
 
+def get_super_call_kwargs_repr(layer: LayerBase) -> str:
```
albertz, line 500: Why do you make this as a separate function, and not a member function of `LayerSignature`?

models/_generate_layers.py
```diff
@@ -246,7 +247,7 @@ def module_call_needs_suppress_shadow_builtin_warning(self):
         return True
     return False
 
-  _IgnoreParamNames = {
+  IgnoreParamNames = {
```
albertz, line 250: This should not be needed when `get_super_call_kwargs_repr` is a member function here.

models/_generate_layers.py
```diff
@@ -81,7 +81,8 @@ def setup():
         if param.is_module_init_arg():
           print(param.get_module_param_docstring(indent="    "), file=f)
       print('    """', file=f)
-      print(f"    super().__init__({'**kwargs' if issubclass(cls_base, LayerBase) else ''})", file=f)
+      params = get_super_call_kwargs_repr(layer_class)
+      print(f"    super().__init__({params})", file=f)
```
albertz, line 85: I think I would refactor the function such that you do directly:
```suggestion
      print(f"    {sig.get_super_call_code_str()}", file=f)
```
This is cleaner, and also makes the documentation of the function `get_super_call_code_str` simpler and cleaner and more clear.

models/_generate_layers.py
albertz: And additionally, refactor and call it `get_super_call_code_str`, as I suggested above.

models/_generate_layers.py
albertz: I think I would just add `batch_norm` to `_IgnoreParamNames`. I'm not sure it is a good idea to anyway use it directly on other layers. We do not need that here (and it might be even better to have it more clean).

models/_generate_layers.py
albertz: Ok, see my comment above. I would just add `batch_norm` to `_IgnoreParamNames`.

models/_generate_layers.py
albertz: (But this will be obsolete when you take my other suggestion about the refactor.)

Atticus1806:
> You still miss the generated layers file here in the PR. (Which would be helpful to see what it actually produces.)

It should be attached in the changed files, is it not?

albertz:
> > You still miss the generated layers file here in the PR. (Which would be helpful to see what it actually produces.)
> 
> It should be attached in the changed files, is it not?

Oh sorry, it is. It just was not displayed ("Large diffs are not rendered by default.").

albertz:
> > > You still miss the generated layers file here in the PR. (Which would be helpful to see what it actually produces.)
> > 
> > 
> > It should be attached in the changed files, is it not?
> 
> Oh sorry, it is. It just was not displayed ("Large diffs are not rendered by default.").

Hm ok, but either GitHub doesn't correctly show the diff (does that work for you? it shows me that every single line was changed) or there are some incorrect changes.

When you look at the changes locally, e.g. via `git log -p`, what do you see? Can you maybe post that here? I would have assumed that there are only minor changes?

Atticus1806:
For the "initial" generation it claims that the whole file is "changed" because of newline on Windows I think 😅 
I will implement your suggested changes and then run it on a different device, then the overall diff should work again.

albertz:
> For the "initial" generation it claims that the whole file is "changed" because of newline on Windows I think 

Ah, this was another potential reason. But I thought that GitHub would show some special symbols for the strange Windows-style newline. But is this really the reason? Did you check the newline-style? And how does that happen? `open(..., "w")` on Windows will automatically do that?

Ok, it seems it does, after reading the [Python doc on `open`](https://docs.python.org/3/library/functions.html#open).

I think you can just add `newline=''` to `open`, and then it should be correct. Can you try to do that?

models/_generate_layers.py
Atticus1806: Adding it to _IgnoreParamNames excludes it from `_Base` layer too right? Or is there an exception for `_Base` somwhere I am missing?

Atticus1806: And I think in `_Base` it is necessary

models/_generate_layers.py
Atticus1806: This would sometimes cause extra whitespaces in the super call, which e.g. would be filtered out by the tests. I was not sure where exactly but from what I found in both key and value and thought it wouldn't harm to just strip them here, but I agree doing that beforehand is better.

models/_generate_layers.py
Atticus1806: Oh yes you are right, my bad. Will be changed

models/_generate_layers.py
```diff
@@ -345,6 +344,42 @@ def _parse_init_docstring(self):
   def __repr__(self):
     return f"<{self.__class__.__name__} {self.layer_class.__name__} {self.inspect_init_sig}>"
 
+  def get_super_call_code_str(self) -> str:
+    """
+    Inspects the given layer in RETURNN, extracts the super call from there, removes unwanted parameters.
+
+    :return: kwargs_repr
+      Python source code of kwargs for the super() call of the layer which are not excluded.
+      Such that `f"super().__init__({kwargs_repr})"` is valid code.
```
albertz, line 353: Now this documentation is again not correct.

models/_generate_layers.py
```diff
@@ -345,6 +344,42 @@ def _parse_init_docstring(self):
...
+    """
+    # get code as string list
+    code = inspect.getsource(self.layer_class.__init__).splitlines()
+
+    # get list of lines starting with super and take the first one
+    lines = [line.strip() for line in code if line.strip().startswith("super")]
+    if not lines:
+      return ''
```
albertz, line 361: Should be:
```suggestion
      return 'super().__init__()'
```

models/_generate_layers.py
```diff
@@ -345,6 +344,42 @@ def _parse_init_docstring(self):
...
+
+    # get list of tuples for parameter with (param_name, value)
+    tup_ls = [x.split("=") for x in call_pruned.split(",") if x.strip() != "**kwargs"]
+    tup_ls = [(key.strip(), value.strip()) for (key, value) in tup_ls]
+    # additionally excluded parameters
+
+    # remove excluded params
+    tup_ls = [f"{key}={value}" for (key, value) in tup_ls if key not in self._IgnoreParamNames]
+    tup_ls += ["**kwargs"]
+    return ", ".join(tup_ls)
```
albertz, line 381:
```suggestion
    return "super().__init__(%s)" % ", ".join(tup_ls)
```

Atticus1806:
Okay you should be able to see the changes correctly now, sorry for that conversion error, I just relized that while Pycharm recognizes the difference does not show it as "diff" when pushing to a remote file.

models/_generate_layers.py
```diff
@@ -345,6 +344,42 @@ def _parse_init_docstring(self):
...
+    assert call.endswith(")")
+    # remove "__init__(" and ") from the call string
+    call_pruned = call[len("__init__("):-len(")")]  # call pruned = "beam_size=beam_size, search=search, **kwargs"
+
+    # get list of tuples for parameter with (param_name, value)
+    tup_ls = [x.split("=") for x in call_pruned.split(",") if x.strip() != "**kwargs"]
+    tup_ls = [(key.strip(), value.strip()) for (key, value) in tup_ls]
+    # additionally excluded parameters
+
+    # remove excluded params
```
albertz, line 378: The first comment is outdated.
And I think the second comment also not really needed.
```suggestion
```

models/_generate_layers.py
```diff
@@ -81,7 +81,7 @@ def setup():
         if param.is_module_init_arg():
           print(param.get_module_param_docstring(indent="    "), file=f)
       print('    """', file=f)
-      print(f"    super().__init__({'**kwargs' if issubclass(cls_base, LayerBase) else ''})", file=f)
+      print(f"    super().__init__({sig.get_super_call_code_str()})", file=f)
```
albertz, line 84:
```suggestion
      print(f"    {sig.get_super_call_code_str()}", file=f)
```
(This is just to have it simple and clean as I suggested before. Also to have it consistent to other code, see other functions `..._code_str()`.)

models/_generate_layers.py
Atticus1806: updated it, is it better now? Or still to vague (not 100% how to formulate the documentation)?

Atticus1806:
I really like the suggestion with parsing the whole call inside the function and not just the params, could have thought of that myself though 😅 

Anyways it should be updated now, as you already predicted only a few layers are actually updated, including Choice and Eval layer.

models/_generate_layers.py
```diff
@@ -344,6 +344,39 @@ def _parse_init_docstring(self):
   def __repr__(self):
     return f"<{self.__class__.__name__} {self.layer_class.__name__} {self.inspect_init_sig}>"
 
+  def get_super_call_code_str(self) -> str:
+    """
+    Inspects the super call of the layer class in RETURNN, extracts the super call parameters from there,
+    removes unwanted parameters and builds a super call which can be written as super call into _generated_layers.py
+    for that class.
+
+    :return: filtered parameters as kwargs_repr in a super call e.g `f"super().__init__({kwargs_repr})"`
```
albertz, line 353: This is wrong.
It returns the code string for the super call.
```suggestion
    :return: sth like `super().__init__(...)`
```

models/_generate_layers.py
albertz: See new comment below.

albertz (approved):

PR #15, opened 2021-08-02
Exclude __class__ from extra_opts

Atticus1806:
This PR adds the check wether the param for extra opts is `__class_` which otherwise would cause it to be written into the dictionary like `'__class__': <class 'returnn_common.models.base.Rec'>`

### DIFF
--- a/models/base.py
+++ b/models/base.py
@@ -285,7 +285,7 @@ class Rec(ISubnet):
     super(Rec, self).__init__()
     self.extra_opts = {
       key: value for (key, value) in locals().items()
-      if value is not NotSpecified and key != "self"}
+      if value is not NotSpecified and key != "self" and key != "__class__"}
 
   def step(self, *args, **kwargs) -> LayerRef:
     """
--- a/models/base.py
+++ b/models/base.py
@@ -285,7 +285,7 @@ class Rec(ISubnet):
     super(Rec, self).__init__()
     self.extra_opts = {
       key: value for (key, value) in locals().items()
-      if value is not NotSpecified and key != "self" and key != "__class__"}
+      if value is not NotSpecified and key not in {"self", "__class__"}}
 
   def step(self, *args, **kwargs) -> LayerRef:
     """


models/base.py
```diff
@@ -285,7 +285,7 @@ def __init__(self, *,
     super(Rec, self).__init__()
     self.extra_opts = {
       key: value for (key, value) in locals().items()
-      if value is not NotSpecified and key != "self"}
+      if value is not NotSpecified and key != "self" and key != "__class__"}
```
albertz, line 288:
```suggestion
      if value is not NotSpecified and key not in {"self", "__class__"}}
```

albertz:
> This PR adds the check wether the param for extra opts is `__class_` which otherwise would cause it to be written into the dictionary like `'__class__': <class 'returnn_common.models.base.Rec'>`

The code looks fine. But From just reading this code diff, I don't quite understand why this can happen.

Atticus1806:
We iterate over locals().items() which also includes `__class__`. Isn't this standard behavior in a class?

Maybe in general it would be even better to check more generally that the variable does not start with `_` instead of having `__class__` as a special case. Something like: `if value is not NotSpecified and key not in {"self"} and not key.startswith("_"}`

albertz:
> We iterate over locals().items()

Ah, I did not realize when just looking at this diff that it is about this.

> which also includes `__class__`. Isn't this standard behavior in a class?

This was not when I executed it. Maybe it was introduced in a later Python version.
 
> Maybe in general it would be even better to check more generally that the variable does not start with `_` instead of having `__class__` as a special case. Something like: `if value is not NotSpecified and key not in {"self"} and not key.startswith("_")`

I'm not sure if this is really always true. Maybe for prefix `__`. But also, I wonder what potential other special local variables could be there. The usage of `locals()` is anyway a bit of a hack and this here is an example where it is a bit unexpected, and maybe some other Python interpreter (PyPy or so) also behave different. I'm not even sure that it is guaranteed that `locals()` really would return all local vars (maybe some get optimized away or so).

But I would just leave it like it is now (and your new change), until it breaks.

albertz (approved):

Issue #52, opened 2021-11-03
Implement standard attention and self-attention module


albertz:
Via `cum_concat` etc.
https://github.com/rwth-i6/returnn/issues/391
https://github.com/rwth-i6/returnn/pull/656
https://github.com/rwth-i6/returnn/pull/589
https://github.com/rwth-i6/returnn/pull/590

albertz:
For this, I think we really should clarify the dim tags first (#17).

albertz:
I implemented some first variants. Although they are not really tested, but probably we want to test them together with #55 or #58.

If anything is broken, or incomplete, or needs to be adopted for #17, I think we should just open new issues.

albertz:
Just for reference, this lacks positional encoding so far.

Some relevant discussion:
https://github.com/pytorch/pytorch/issues/24826
https://github.com/allenai/allennlp/issues/3398

albertz:
See also: [Pytorch `MultiheadAttention`](https://github.com/pytorch/pytorch/blob/9fb3ba9d7bc5f6f8852a2bafe053b16a26d73e51/torch/nn/modules/activation.py#L862).

PR #55, opened 2021-11-03
Transformer Modules

Atticus1806:
Fix #53.

This is a draft for now, since Attention Modules from #52 need to be implemented for that. Comments on code style, changes etc. are welcome already of course. 

Orienting on PyTorch for naming of variables and functions.
[PyTorch documentation](https://pytorch.org/docs/stable/nn.html#transformer-layers).

Note that PyTorch also has a masking logic in almost all modules, which was left out for now, since it is used in the Attention Modules in PyTorch. Depending on the structure of `rc` Attention this can of course be added again.

Issue #53, opened 2021-11-03
Implement standard Transformer encoder and decoder


albertz:
Very similar to the PyTorch modules.
Based on attention and self-attention from #52.

Issue #17, opened 2021-08-05
Design/Handling of dimension tags


albertz:
Like batch dim, spatial dims (with dynamic lengths), or static dims (named, or also unnamed).

`DimensionTag` in RETURNN. Directly use that, or wrap it somehow?

Relevant for all layers which define some shape or dimension (e.g. `Const`, `Variable`).

Atticus1806:
Would this replace the possibility to define `out_type` (which you would do in RETURNN)? I didn't continue with the debugging yet, but I think I got some kind of related issue in my config right now. Cannot tell for sure, will report back once I had time to check.

albertz:
I'm not sure yet whether this would at the same time remove the old way, or just be an additional way.

In any case, I don't think this would really change much in behavior, depending on what you did. So whatever problem you had, you might have it again now. But not sure about the details on your issue. You maybe should report a separate issue about it. But also not sure if this is even related here to returnn-common, or just RETURNN itself?

But also, in general, you never should use `out_type` anyway. If you do, this sounds like you probably should do it in a different way.

The dimension tags would be used whenever you must specify an axis (for `Reduce`, `Dot`, etc), or a new shape (for `Const`, `Variable`, etc).

In general, I want to push the concept of named tensors (named array axes; i.e. dimension tags in RETURNN) much more. I want to deprecate or even not allow axis specified as integers. Axis specified as integers (index) can very easily become a problem because the index can easily change (because we optimize sth differently, then the order of axis will be different, or the user adds some other layers and doesn't notice, etc). So the order of axes should not matter at no point.

Atticus1806:
>But also, in general, you never should use out_type anyway. If you do, this sounds like you probably should do it in a different way.

Without further investigation on my own config, how would this work with an `Eval` layer here? Is returnn able to (or should be able to) implicitly solve the eval and handle the dimension? 

>In general, I want to push the concept of named tensors (named array axes; i.e. dimension tags in RETURNN) much more.

I like this idea, this would make things a lot more straight forward too, I think.

albertz:
> > But also, in general, you never should use out_type anyway. If you do, this sounds like you probably should do it in a different way.
> 
> Without further investigation on my own config, how would this work with an `Eval` layer here? Is returnn able to (or should be able to) implicitly solve the eval and handle the dimension?

[...]

PR #58, opened 2021-11-04
Conformer Modules

### LINKED ISSUE #54: Implement standard Conformer encoder
Similar as Transformer in #53.
Using self-attention from #52.
Maybe making other sensible building blocks as needed.

mmz33:
Fix #54.

This is a draft now. It still requires the implementation of MultiHeadSelfAttention from #52.

### DIFF
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -1,0 +1,204 @@
+"""
+Conformer code.
+Ref: https://arxiv.org/abs/2005.08100
+"""
+
+from typing import Tuple, List, Union
+from . import Module, ModuleList, LayerRef, Linear, dropout, layer_norm, batch_norm, Conv, swish, glu, split_dims, \
+    merge_dims, pool
+
+
+class _PositionwiseFeedForward(Module):
+    """
+    Conformer position-wise feedforward neural network layer
+        FF -> Activation -> Dropout -> FF
+    """
+
+    def __init__(self, d_model: int, d_ff: int, dropout: float, activation, l2: float = 0.0):
+        """
+        :param d_model:
+        :param d_ff:
+        :param dropout:
+        :param activation:
+        :param l2:
+        """
+        super().__init__()
+
+        self.dropout = dropout
+        self.activation = activation
+
+        self.linear1 = Linear(n_out=d_ff, l2=l2)
+        self.linear2 = Linear(n_out=d_model, l2=l2)
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        return self.linear2(dropout(self.activation(self.linear1(inp)), dropout=self.dropout))
+
+
+class _ConformerConvBlock(Module):
+    """
+    Conformer convolution block
+        FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+    """
+
+    def __init__(self, d_model: int, kernel_size: Tuple[int], l2: float = 0.0):
+        """
+        :param d_model:
+        :param kernel_size:
+        :param l2:
+        """
+        super().__init__()
+
+        self.positionwise_conv1 = Linear(n_out=d_model * 2, l2=l2)
+        self.depthwise_conv = Conv(n_out=d_model, filter_size=kernel_size, groups=d_model, l2=l2, padding='same')
+        self.positionwise_conv2 = Linear(n_out=d_model, l2=l2)
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        x_conv1 = self.positionwise_conv1(inp)
+        x_act = glu(x_conv1)
+        x_depthwise_conv = self.depthwise_conv(x_act)
+        x_bn = batch_norm(x_depthwise_conv)
+        x_swish = swish(x_bn)
+        x_conv2 = self.positionwise_conv2(x_swish)
+        return x_conv2
+
+
+class _ConformerConvSubsampleLayer(Module):
+    """
+    Conv 2D block with optional max-pooling
+    """
+
+    def __init__(self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
+                 channel_sizes: List[int], l2: float = 0.0, dropout: float = 0.3, act: str = 'relu',
+                 padding: str = 'same'):
+        """
+        :param filter_sizes:
+        :param pool_sizes:
+        :param channel_sizes:
+        :param l2:
+        :param dropout:
+        :param act:
+        :param padding:
+        """
+        super().__init__()
+
+        self.dropout = dropout
+        self.pool_sizes = pool_sizes
+
+        self.conv_layers = ModuleList()
+        for filter_size, channel_size in zip(filter_sizes, channel_sizes):
+            self.conv_layers.append(
+                Conv(l2=l2, activation=act, filter_size=filter_size, n_out=channel_size, padding=padding))
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        x = split_dims(inp, axis='F', dims=(-1, 1))
+        for i, conv_layer in enumerate(self.conv_layers):
+            x = conv_layer(x)
+            if self.pool_sizes and i < len(self.pool_sizes):
+                x = pool(x, pool_size=self.pool_sizes[i], padding='same', mode='max')
+            if self.dropout:
+                x = dropout(x, dropout=self.dropout)
+        out = merge_dims(x, axes='static')
+        return out
+
+
+class ConformerEncoderLayer(Module):
+    """
+    Represents a conformer block
+    """
+
+    def __init__(self, conv_kernel_size: Tuple[int], ff_act, ff_dim: int, dropout: float, att_dropout: float,
+                 enc_key_dim: int, att_n_heads: int, l2: float):
+        """
+        :param conv_kernel_size:
+        :param ff_act:
+        :param ff_dim:
+        :param dropout:
+        :param att_dropout:
+        :param enc_key_dim:
+        :param att_n_heads:
+        :param l2:
+        """
+        super().__init__()
+
+        self.dropout = dropout
+
+        self.ffn1 = _PositionwiseFeedForward(
+            d_model=enc_key_dim, d_ff=ff_dim, dropout=dropout, activation=ff_act, l2=l2)
+
+        self.ffn2 = _PositionwiseFeedForward(
+            d_model=enc_key_dim, d_ff=ff_dim, dropout=dropout, activation=ff_act, l2=l2)
+
+        self.conv_module = _ConformerConvBlock(d_model=enc_key_dim, kernel_size=conv_kernel_size)
+
+        self.mhsa_module = MultiheadAttention(d_model, att_n_heads, dropout=att_dropout)  # TODO: to be implemented
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        # FFN
+        x_ffn1_ln = layer_norm(inp)
+        x_ffn1 = self.ffn1(x_ffn1_ln)
+        x_ffn1_out = 0.5 * dropout(x_ffn1, dropout=self.dropout) + inp
+
+        # MHSA
+        x_mhsa_ln = layer_norm(x_ffn1_out)
+        x_mhsa = self.mhsa_module(x_mhsa_ln)
+        x_mhsa_out = x_mhsa + x_ffn1_out
+
+        # Conv
+        x_conv_ln = layer_norm(x_mhsa_out)
+        x_conv = self.conv_module(x_conv_ln)
+        x_conv_out = dropout(x_conv, dropout=self.dropout) + x_mhsa_out
+
+        # FFN
+        x_ffn2_ln = layer_norm(x_conv_out)
+        x_ffn2 = self.ffn2(x_ffn2_ln)
+        x_ffn2_out = 0.5 * dropout(x_ffn2, dropout=self.dropout) + x_conv_out
+
+        # last LN layer
+        return layer_norm(x_ffn2_out)
+
+
+class ConformerEncoder(Module):
+    """
+    Represents Conformer encoder architecture
+    """
+
+    def __init__(self, encoder_layer: Module, num_blocks: int, conv_kernel_size: Tuple[int, ...] = (32,), ff_act=swish,
+                 ff_dim: int = 512, dropout: float = 0.1, att_dropout: float = 0.1, enc_key_dim: int = 256,
+                 att_n_heads: int = 4, l2: float = 0.0):
+        """
+        :param encoder_layer:
+        :param num_blocks:
+        :param conv_kernel_size:
+        :param ff_act:
+        :param ff_dim:
+        :param dropout:
+        :param att_dropout:
+        :param enc_key_dim:
+        :param att_n_heads:
+        :param l2:
+        """
+        super().__init__()
+
+        self.dropout = dropout
+
+        self.conv_subsample_layer = _ConformerConvSubsampleLayer(
+            filter_sizes=[(3, 3), (3, 3)], pool_sizes=[(2, 2), (2, 2)], channel_sizes=[enc_key_dim, enc_key_dim],
+            l2=l2, dropout=dropout)
+
+        self.linear = Linear(n_out=enc_key_dim, l2=l2, with_bias=False)
+
+        self.conformer_blocks = ModuleList([
+            encoder_layer(
+                conv_kernel_size=conv_kernel_size, ff_act=ff_act, ff_dim=ff_dim, dropout=dropout,
+                att_dropout=att_dropout, enc_key_dim=enc_key_dim, att_n_heads=att_n_heads, l2=l2
+            )
+            for _ in range(num_blocks)
+        ])
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        x_subsample = self.conv_subsample_layer(inp)
+        x_linear = self.linear(x_subsample)
+        x = dropout(x_linear, dropout=self.dropout)
+        for conformer_block in self.conformer_blocks:
+            x = conformer_block(x)
+        return x
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -79,9 +79,9 @@ class _ConformerConvSubsample(nn.Module):
   """
 
   def __init__(
-        self, filter_sizes: List[Tuple[int, int]], pool_sizes: Optional[List[Tuple[int, int]]],
-        channel_sizes: List[int], dropout: float = 0.3, activation: Callable[[nn.LayerRef], nn.LayerRef] = nn.relu,
-        padding: str = 'same'):
+        self, filter_sizes: List[Tuple[int, int]], channel_sizes: List[int],
+        pool_sizes: Optional[List[Tuple[int, int]]] = None, dropout: float = 0.3,
+        activation: Callable[[nn.LayerRef], nn.LayerRef] = nn.relu, padding: str = 'same'):
     """
     :param filter_sizes:
     :param pool_sizes:
@@ -122,7 +122,7 @@ class ConformerEncoderLayer(nn.Module):
 
   def __init__(
         self, conv_kernel_size: int, activation_ff: Callable[[nn.LayerRef], nn.LayerRef], dim_ff: int, dropout: float,
-        att_dropout: float, enc_key_dim: int, num_heads: int, batch_norm_opts: Optional[Dict[str, Any]]):
+        att_dropout: float, enc_key_dim: int, num_heads: int, batch_norm_opts: Optional[Dict[str, Any]] = None):
     """
     :param conv_kernel_size:
     :param activation_ff:
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -206,10 +206,8 @@ class ConformerEncoder(nn.Module):
     self.conformer_blocks = nn.Sequential(
       encoder_layer(
         conv_kernel_size=conv_kernel_size, activation_ff=activation_ff, dim_ff=dim_ff, dropout=dropout,
-        att_dropout=att_dropout, enc_key_dim=enc_key_dim, num_heads=num_heads, batch_norm_opts=batch_norm_opts
-      )
-      for _ in range(num_blocks)
-    )
+        att_dropout=att_dropout, enc_key_dim=enc_key_dim, num_heads=num_heads, batch_norm_opts=batch_norm_opts)
+      for _ in range(num_blocks))
 
   def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
     x_subsample = self.conv_subsample_layer(inp)
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -78,7 +78,7 @@ class _ConformerConvSubsampleLayer(nn.Module):
 
   def __init__(
       self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
-      channel_sizes: List[int], dropout: float = 0.3, activation: str = 'relu', padding: str = 'same'):
+      channel_sizes: List[int], dropout: float = 0.3, activation=nn.relu, padding: str = 'same'):
     """
     :param filter_sizes:
     :param pool_sizes:
@@ -91,17 +91,19 @@ class _ConformerConvSubsampleLayer(nn.Module):
 
     self.dropout = dropout
     self.pool_sizes = pool_sizes
+    self.activation = activation
 
     self.conv_layers = nn.ModuleList()
     assert len(filter_sizes) == len(channel_sizes)
     for filter_size, channel_size in zip(filter_sizes, channel_sizes):
       self.conv_layers.append(
-        nn.Conv(activation=activation, filter_size=filter_size, n_out=channel_size, padding=padding))
+        nn.Conv(filter_size=filter_size, n_out=channel_size, padding=padding))
 
   def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
     x = nn.split_dims(inp, axis='F', dims=(-1, 1))
     for i, conv_layer in enumerate(self.conv_layers):
       x = conv_layer(x)
+      x = self.activation(x)
       if self.pool_sizes and i < len(self.pool_sizes):
         x = nn.pool(x, pool_size=self.pool_sizes[i], padding='same', mode='max')
       if self.dropout:
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -131,7 +131,7 @@ class ConformerEncoderLayer(nn.Module):
     :param att_dropout: attention dropout value
     :param out_dim: the output feature dimension
     :param num_heads: the number of attention heads
-    :param batch_norm_opts: batch norm options
+    :param batch_norm_opts: passed to :class:`nn.BatchNorm`
     """
     super().__init__()
 
@@ -181,7 +181,7 @@ class ConformerEncoder(nn.Module):
 
   def __init__(self, encoder_layer: ConformerEncoderLayer, num_layers: int):
     """
-    :param encoder_layer: an instance of `class:ConformerEncoderLayer`
+    :param encoder_layer: an instance of :class:`ConformerEncoderLayer`
     :param num_layers: the number of encoder layers
     """
     super().__init__()
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -31,7 +31,11 @@ class _PositionwiseFeedForward(nn.Module):
     self.linear2 = nn.Linear(n_out=dim_model, l2=l2)
 
   def forward(self, inp: LayerRef) -> LayerRef:
-    return self.linear2(nn.dropout(self.activation(self.linear1(inp)), dropout=self.dropout))
+    x_ff1 = self.linear1(inp)
+    x_act = self.activation(x_ff1)
+    x_drop = nn.dropout(x_act, dropout=self.dropout)
+    x_ff2 = self.linear2(x_drop)
+    return x_ff2
 
 
 class _ConformerConvBlock(nn.Module):
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -178,8 +178,7 @@ class ConformerEncoder(nn.Module):
   Represents Conformer encoder architecture
   """
 
-  def __init__(
-        self, encoder_layer: Union[ConformerEncoderLayer, Any], num_blocks: int):
+  def __init__(self, encoder_layer: Union[ConformerEncoderLayer, Any], num_blocks: int):
     """
     :param encoder_layer:
     :param num_blocks:
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -3,7 +3,7 @@ Conformer code.
 Ref: https://arxiv.org/abs/2005.08100
 """
 
-from typing import Tuple, List, Union
+from typing import Tuple, List, Union, Callable
 from .. import nn
 
 
@@ -13,7 +13,7 @@ class _PositionwiseFeedForward(nn.Module):
       FF -> Activation -> Dropout -> FF
   """
 
-  def __init__(self, out_dim: int, dim_ff: int, dropout: float, activation):
+  def __init__(self, out_dim: int, dim_ff: int, dropout: float, activation: Callable[[nn.LayerRef], nn.LayerRef]):
     """
     :param out_dim:
     :param dim_ff:
@@ -78,7 +78,8 @@ class _ConformerConvSubsampleLayer(nn.Module):
 
   def __init__(
       self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
-      channel_sizes: List[int], dropout: float = 0.3, activation=nn.relu, padding: str = 'same'):
+      channel_sizes: List[int], dropout: float = 0.3, activation: Callable[[nn.LayerRef], nn.LayerRef] = nn.relu,
+      padding: str = 'same'):
     """
     :param filter_sizes:
     :param pool_sizes:
@@ -118,8 +119,8 @@ class ConformerEncoderLayer(nn.Module):
   """
 
   def __init__(
-      self, conv_kernel_size: int, activation_ff, dim_ff: int, dropout: float, att_dropout: float, enc_key_dim: int,
-      num_heads: int):
+      self, conv_kernel_size: int, activation_ff: Callable[[nn.LayerRef], nn.LayerRef], dim_ff: int, dropout: float,
+      att_dropout: float, enc_key_dim: int, num_heads: int):
     """
     :param conv_kernel_size:
     :param activation_ff:
@@ -174,8 +175,9 @@ class ConformerEncoder(nn.Module):
   """
 
   def __init__(
-      self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size: int = 32, activation_ff=nn.swish,
-      dim_ff: int = 512, dropout: float = 0.1, att_dropout: float = 0.1, enc_key_dim: int = 256, num_heads: int = 4):
+      self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size: int = 32,
+      activation_ff: Callable[[nn.LayerRef], nn.LayerRef] = nn.swish, dim_ff: int = 512, dropout: float = 0.1,
+      att_dropout: float = 0.1, enc_key_dim: int = 256, num_heads: int = 4):
     """
     :param encoder_layer:
     :param num_blocks:
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -187,7 +187,7 @@ class ConformerEncoder(nn.Module):
 
     self.linear = nn.Linear(n_out=enc_key_dim, l2=l2, with_bias=False)
 
-    self.conformer_blocks = nn.ModuleList([
+    self.conformer_blocks = nn.Sequential([
       encoder_layer(
         conv_kernel_size=conv_kernel_size, ff_act=ff_act, ff_dim=ff_dim, dropout=dropout,
         att_dropout=att_dropout, enc_key_dim=enc_key_dim, att_n_heads=att_n_heads, l2=l2
@@ -199,6 +199,5 @@ class ConformerEncoder(nn.Module):
     x_subsample = self.conv_subsample_layer(inp)
     x_linear = self.linear(x_subsample)
     x = nn.dropout(x_linear, dropout=self.dropout)
-    for conformer_block in self.conformer_blocks:
-      x = conformer_block(x)
+    x = self.conformer_blocks(x)
     return x
--- a/nn/conformer.py
+++ b/nn/conformer.py
@@ -3,7 +3,7 @@ Conformer code.
 Ref: https://arxiv.org/abs/2005.08100
 """
 
-from typing import Tuple, List, Callable, Optional, Dict, Any
+from typing import Tuple, List, Callable, Optional, Dict, Any, Union
 from .. import nn
 import copy
 
@@ -54,14 +54,12 @@ class _ConformerConvBlock(nn.Module):
     self.depthwise_conv = nn.Conv(n_out=out_dim, filter_size=(kernel_size,), groups=out_dim, padding='same')
     self.positionwise_conv2 = nn.Linear(n_out=out_dim)
 
-    batch_norm_opts = copy.deepcopy(batch_norm_opts)
     if batch_norm_opts is None:
       batch_norm_opts = {}
-    epsilon = batch_norm_opts.pop('epsilon', 1e-5)
-    momentum = batch_norm_opts.pop('momentum', 0.1)
-    self.batch_norm = nn.BatchNorm(
-      epsilon=epsilon, momentum=momentum, update_sample_only_in_training=True, delay_sample_update=True,
-      **batch_norm_opts)
+    batch_norm_opts = batch_norm_opts.copy()
+    batch_norm_opts.setdefault('epsilon', 1e-5)
+    batch_norm_opts.setdefault('momentum', 0.1)
+    self.batch_norm = nn.BatchNorm(update_sample_only_in_training=True, delay_sample_update=True, **batch_norm_opts)
 
   def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
     x_conv1 = self.positionwise_conv1(inp)
@@ -84,9 +82,9 @@ class _ConformerConvSubsample(nn.Module):
         padding: str = 'same'):
     """
     :param filter_sizes:
-    :param pool_sizes:
     :param channel_sizes:
     :param dropout:
+    :param pool_sizes:
     :param activation:
     :param padding:
     """
@@ -121,29 +119,32 @@ class ConformerEncoderLayer(nn.Module):
   """
 
   def __init__(
-        self, conv_kernel_size: int, activation_ff: Callable[[nn.LayerRef], nn.LayerRef], dim_ff: int, dropout: float,
-        att_dropout: float, enc_key_dim: int, num_heads: int, batch_norm_opts: Optional[Dict[str, Any]] = None):
+        self, conv_kernel_size: int = 32, activation_ff: Callable[[nn.LayerRef], nn.LayerRef] = nn.swish,
+        dim_ff: int = 2048, dropout: float = 0.1, att_dropout: float = 0.1, out_dim: int = 512, num_heads: int = 8,
+        batch_norm_opts: Optional[Dict[str, Any]] = None):
     """
     :param conv_kernel_size:
     :param activation_ff:
-    :param ff_dim:
+    :param dim_ff:
     :param dropout:
     :param att_dropout:
-    :param enc_key_dim:
+    :param out_dim:
     :param num_heads:
+    :param batch_norm_opts:
     """
     super().__init__()
 
     self.dropout = dropout
+    self.out_dim = out_dim
 
     self.ffn1 = _PositionwiseFeedForward(
-      out_dim=enc_key_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff)
+      out_dim=out_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff)
 
     self.ffn2 = _PositionwiseFeedForward(
-      out_dim=enc_key_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff)
+      out_dim=out_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff)
 
     self.conv_block = _ConformerConvBlock(
-      out_dim=enc_key_dim, kernel_size=conv_kernel_size, batch_norm_opts=batch_norm_opts)
+      out_dim=out_dim, kernel_size=conv_kernel_size, batch_norm_opts=batch_norm_opts)
 
     self.self_att = MultiheadAttention(enc_key_dim, num_heads, dropout=att_dropout)  # TODO: to be implemented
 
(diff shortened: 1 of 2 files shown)


mmz33:
The parameters names still need to be changed to be consistent with #53.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+Ref: https://arxiv.org/abs/2005.08100
+"""
+
+from typing import Tuple, List, Union
+from . import Module, ModuleList, LayerRef, Linear, dropout, layer_norm, batch_norm, Conv, swish, glu, split_dims, \
+    merge_dims, pool
+
+
+class _PositionwiseFeedForward(Module):
+    """
```
albertz, line 12: Indentation should be consistent (2 spaces).

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
+"""
+Conformer code.
+Ref: https://arxiv.org/abs/2005.08100
+"""
+
+from typing import Tuple, List, Union
+from . import Module, ModuleList, LayerRef, Linear, dropout, layer_norm, batch_norm, Conv, swish, glu, split_dims, \
+    merge_dims, pool
```
albertz, line 8: Similar argument as in #55, I think some import like this is cleaner:
```suggestion
from .. import nn
```

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class _PositionwiseFeedForward(Module):
+    """
+    Conformer position-wise feedforward neural network layer
+        FF -> Activation -> Dropout -> FF
+    """
+
+    def __init__(self, d_model: int, d_ff: int, dropout: float, activation, l2: float = 0.0):
```
albertz, line 17: Similar as in #55, we should make it more consistent.

albertz: I wonder whether we should handle options like `l2` in a more generic way...

nn/conformer.py
mmz33: ah yes. I will configure my PyCharm settings to be consistent.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+        return self.linear2(dropout(self.activation(self.linear1(inp)), dropout=self.dropout))
+
+
+class _ConformerConvBlock(Module):
+    """
+    Conformer convolution block
+        FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+    """
+
+    def __init__(self, d_model: int, kernel_size: Tuple[int], l2: float = 0.0):
```
albertz, line 43: Maybe make `kernel_size: int` here and then pass on `[kernel_size]` below.

albertz: Also consistent arg naming.

albertz: And `l2` (#59).

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+        self.dropout = dropout
+        self.pool_sizes = pool_sizes
+
+        self.conv_layers = ModuleList()
+        for filter_size, channel_size in zip(filter_sizes, channel_sizes):
+            self.conv_layers.append(
+                Conv(l2=l2, activation=act, filter_size=filter_size, n_out=channel_size, padding=padding))
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        x = split_dims(inp, axis='F', dims=(-1, 1))
```
albertz, line 93: This will likely look different when we use #17.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self.dropout = dropout
+
+        self.conv_subsample_layer = _ConformerConvSubsampleLayer(
+            filter_sizes=[(3, 3), (3, 3)], pool_sizes=[(2, 2), (2, 2)], channel_sizes=[enc_key_dim, enc_key_dim],
+            l2=l2, dropout=dropout)
+
+        self.linear = Linear(n_out=enc_key_dim, l2=l2, with_bias=False)
+
+        self.conformer_blocks = ModuleList([
```
albertz, line 190: `Sequential` might be a better fit, right?

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+            for _ in range(num_blocks)
+        ])
+
+    def forward(self, inp: LayerRef) -> LayerRef:
+        x_subsample = self.conv_subsample_layer(inp)
+        x_linear = self.linear(x_subsample)
+        x = dropout(x_linear, dropout=self.dropout)
+        for conformer_block in self.conformer_blocks:
+            x = conformer_block(x)
```
albertz: When you make it `Sequential` above, then this is simply:
```suggestion
        x = self.conformer_blocks(x)
```

nn/math_.py
```diff
@@ -50,3 +50,8 @@ def sigmoid(x: LayerRef) -> Layer:
 def swish(x: LayerRef) -> Layer:
   """swish"""
   return activation(x, activation="swish")
+
+
+def glu(x: LayerRef) -> Layer:
+  """GLU"""
+  return activation(x, activation='glu')
```
albertz, line 57: Make the order a bit nicer, put it after `gelu`. Also add a link to the paper.

nn/conformer.py
```diff
@@ -0,0 +1,205 @@
+"""
+Conformer code.
+Ref: https://arxiv.org/abs/2005.08100
+"""
+
+from typing import Tuple, List, Union
+
+from . import Module, ModuleList, LayerRef, Linear, dropout, layer_norm, batch_norm, Conv, swish, glu, split_dims, \
+  merge_dims, pool
```
albertz, line 9:
```suggestion
from .. import nn
```

albertz:
See also my other comments (they don't show up in the "Files changed" view because they got outdated due to the changed indent). Mark them as "resolved" when you resolved them.

albertz:
Btw, as usual, also see failing tests.

nn/math_.py
```diff
@@ -27,6 +27,11 @@ def gelu(x: LayerRef) -> Layer:
   return activation(x, activation="gelu")
 
 
+def glu(x: LayerRef) -> Layer:
+  """GLU https://arxiv.org/abs/1612.08083"""
+  return activation(x, activation='glu')
```
albertz, line 32: I just realize what this is. We don't have that on RETURNN side, and also, it would not be via the `ActivationLayer` because this would change the output dimension.

You could implement this here by hand using `nn.split`.

nn/conformer.py
```diff
@@ -0,0 +1,205 @@
...
+
+  def __init__(self, d_model: int, kernel_size: Tuple[int], l2: float = 0.0):
+    """
+    :param d_model:
+    :param kernel_size:
+    :param l2:
+    """
+    super().__init__()
+
+    self.positionwise_conv1 = Linear(n_out=d_model * 2, l2=l2)
```
albertz, line 52: I wonder, as you do GLU, whether it is maybe easier to use:
```suggestion
    self.positionwise_conv1a = Linear(n_out=d_model, l2=l2)
    self.positionwise_conv1b = Linear(n_out=d_model, l2=l2)
```

nn/conformer.py
```diff
@@ -0,0 +1,205 @@
...
+    self.depthwise_conv = Conv(n_out=d_model, filter_size=kernel_size, groups=d_model, l2=l2, padding='same')
+    self.positionwise_conv2 = Linear(n_out=d_model, l2=l2)
+
+  def forward(self, inp: LayerRef) -> LayerRef:
+    x_conv1 = self.positionwise_conv1(inp)
+    x_act = glu(x_conv1)
```
albertz, line 58: And then here maybe:
```
x_conv1a, x_conv1b = self.positionwise_conv1a(inp), self.positionwise_conv1b(inp)
x_act = x_conv1a * nn.sigmoid(x_conv1b)  # GLU
```

nn/conformer.py
```diff
@@ -0,0 +1,205 @@
...
+    x_depthwise_conv = self.depthwise_conv(x_act)
+    x_bn = batch_norm(x_depthwise_conv)
```
albertz: I just realized that `batch_norm` was buggy. It has variables. So you should use the module to make the variable naming consistent.

nn/conformer.py
```diff
@@ -4,12 +4,11 @@
 """
 
 from typing import Tuple, List, Union
+from .. import nn
+from . import LayerRef
```
albertz, line 8: Do you think it is a good idea to have `LayerRef` as an explicit import? Why not just use `nn.LayerRef`?

mmz33: I just thought that I can save some space when defining the parameters. But I can just use `nn.LayerRef`

nn/conformer.py
```diff
@@ -187,7 +187,7 @@ def __init__(self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size:
 
     self.linear = nn.Linear(n_out=enc_key_dim, l2=l2, with_bias=False)
 
-    self.conformer_blocks = nn.ModuleList([
+    self.conformer_blocks = nn.Sequential([
```
albertz, line 190: For `Sequential`, you don't need to pass it as a `list`, you can directly pass it as the arguments.
```suggestion
    self.conformer_blocks = nn.Sequential(
```

albertz: Although, this is slightly different, it will be an iterable here, but anyway, this should also work. If it does not, we should fix `Sequential` such that it works.

nn/math_.py
mmz33: hmm yes it is wrong to use activation class. However, we have gating in RETURNN. It is the `GatingLayer`. But I just noticed that it is in the blacklist here. PyTorch supports it as [here](https://pytorch.org/cppdocs/api/function_namespacetorch_1_1nn_1_1functional_1a789cc6940523640f299df85a3dc5a0fd.html). [Here](https://github.com/espnet/espnet/blob/7a1329a6fb713a17f5860ea7aed2315ade8d9860/espnet/nets/pytorch_backend/conformer/convolution.py#L71) an example how it is used in ESPNet.

albertz: I blacklisted it because it is a bit too less generic (on RETURNN side) and it would be very trivial to reimplement it anyway in a generic way here on returnn-common side. Sth like:
```
def glu(x):
  a, b = nn.split(x, 2)
  return a * nn.sigmoid(b)
```

nn/conformer.py
albertz: Yes use `nn.LayerRef`. This will also save you this line of code here.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+from . import LayerRef
+
+
+class _PositionwiseFeedForward(nn.Module):
+  """
+  Conformer position-wise feedforward neural network layer
+      FF -> Activation -> Dropout -> FF
+  """
+
+  def __init__(self, dim_model: int, dim_ff: int, dropout: float, activation, l2: float = 0.0):
```
albertz, line 17: Remove the `l2`. We should do this in a better more generic way (#59).

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    super().__init__()
+
+    self.positionwise_conv1 = nn.Linear(n_out=dim_model * 2, l2=l2)
+    self.depthwise_conv = nn.Conv(n_out=dim_model, filter_size=kernel_size, groups=dim_model, l2=l2, padding='same')
+    self.positionwise_conv2 = nn.Linear(n_out=dim_model, l2=l2)
+
+    if batch_norm_other_opts is None:
+      batch_norm_other_opts = {}
+
+    self.batch_norm = nn.BatchNorm(epsilon=batch_norm_eps, momentum=batch_norm_momentum,
```
albertz, line 60: If you anyway need to continue the line with a function call, do not add any arguments in this line but only in the following lines.
```suggestion
    self.batch_norm = nn.BatchNorm(
      epsilon=batch_norm_eps, momentum=batch_norm_momentum,
```

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+      update_sample_only_in_training=batch_norm_update_sample_only_in_training,
+      delay_sample_update=batch_norm_delay_sample_update, **batch_norm_other_opts)
+
+  @staticmethod
+  def _glu(v: LayerRef):
```
albertz, line 65: Don't put this here but add it to `math_.py` to the other functions.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    a, b = nn.split(v, axis='F')
```
albertz, line 66: I think it misses the option `2`?

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+      )
+      for _ in range(num_blocks)
+    ])
+
+  def forward(self, inp: LayerRef) -> LayerRef:
+    x_subsample = self.conv_subsample_layer(inp)
+    x_linear = self.linear(x_subsample)
+    x = nn.dropout(x_linear, dropout=self.dropout)
+    x = self.conformer_blocks(x)
+    return x
```
albertz, line 217: Missing newline at the end of file.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    super().__init__()
+
+    self.dropout = dropout
+    self.activation = activation
+
+    self.linear1 = nn.Linear(n_out=dim_ff, l2=l2)
+    self.linear2 = nn.Linear(n_out=dim_model, l2=l2)
+
+  def forward(self, inp: LayerRef) -> LayerRef:
+    return self.linear2(nn.dropout(self.activation(self.linear1(inp)), dropout=self.dropout))
```
albertz, line 34: See my comment on Transformer on this. I prefer it as separate calls.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(self, dim_model: int, kernel_size: int, l2: float = 0.0, batch_norm_eps: float = 1e-5,
+      batch_norm_momentum: float = 0.1, batch_norm_update_sample_only_in_training=True,
+      batch_norm_delay_sample_update=True, batch_norm_other_opts=None):
```
albertz, line 45: Why do you have batch norm custom default options here? Why not just a single `batch_norm_other_opts`?

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    return self.linear2(nn.dropout(self.activation(self.linear1(inp)), dropout=self.dropout))
+
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(self, dim_model: int, kernel_size: int, l2: float = 0.0, batch_norm_eps: float = 1e-5,
```
albertz, line 43: Remove the `l2`.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    super().__init__()
+
+    self.positionwise_conv1 = nn.Linear(n_out=dim_model * 2, l2=l2)
+    self.depthwise_conv = nn.Conv(n_out=dim_model, filter_size=(kernel_size,), groups=dim_model, l2=l2, padding='same')
+    self.positionwise_conv2 = nn.Linear(n_out=dim_model, l2=l2)
+
+    if batch_norm_other_opts is None:
+      batch_norm_other_opts = {}
+
```
albertz, line 59: Remove this empty line.

nn/conformer.py
mmz33: so here first i want to remove `batch_norm_update_sample_only_in_training=True, batch_norm_delay_sample_update=True`. Because these should be True always otherwise batch norm logic is wrong. 
I thought it is nice to see the default params while calling this module as these two params will be the ones mainly tuned just in case.

albertz: The wrong defaults should be fixed directly in the `BatchNorm` module, not here.

albertz: Hm, I'm unsure if it is a good idea to have `batch_norm_momentum` and `batch_norm_eps` here...

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+    self.dropout = dropout
+
+    self.ffn1 = _PositionwiseFeedForward(
+      dim_model=enc_key_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff, l2=l2)
+
+    self.ffn2 = _PositionwiseFeedForward(
+      dim_model=enc_key_dim, dim_ff=dim_ff, dropout=dropout, activation=activation_ff, l2=l2)
+
+    self.conv_module = _ConformerConvBlock(dim_model=enc_key_dim, kernel_size=conv_kernel_size)
```
albertz, line 147: I think I would prefer a better attribute name. You don't need a `_module` postfix for all the modules (and you also don't have that consistently). Maybe just remove the `_module` postfix. Or maybe make it more consistent to the actual module name, e.g. like `conv_block`.

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+
+    self.mhsa_module = MultiheadAttention(enc_key_dim, num_heads, dropout=att_dropout)  # TODO: to be implemented
```
albertz, line 149: Same here on the attrib name. Maybe just `mhsa`. Or maybe better `self_att` (I don't think the multi-head aspect is so relevant to put it into the attrib name).

nn/conformer.py
albertz: You still have the `[`.

nn/conformer.py
```diff
@@ -14,21 +14,20 @@ class _PositionwiseFeedForward(nn.Module):
...
     """
     super().__init__()
 
     self.dropout = dropout
     self.activation = activation
 
-    self.linear1 = nn.Linear(n_out=dim_ff, l2=l2)
-    self.linear2 = nn.Linear(n_out=dim_model, l2=l2)
+    self.linear1 = nn.Linear(n_out=dim_ff)
+    self.linear2 = nn.Linear(n_out=dim_model)
```
albertz, line 29: See my comment on Transformer on this. I think there could be better names.
```suggestion
    self.linear_ff = nn.Linear(n_out=dim_ff)
    self.linear_out = nn.Linear(n_out=dim_model)
```

nn/conformer.py
```diff
@@ -14,21 +14,20 @@ class _PositionwiseFeedForward(nn.Module):
       FF -> Activation -> Dropout -> FF
   """
 
-  def __init__(self, dim_model: int, dim_ff: int, dropout: float, activation, l2: float = 0.0):
+  def __init__(self, dim_model: int, dim_ff: int, dropout: float, activation):
```
albertz, line 16: See my comments on Transformer on the naming of `dim_model`. I think `out_dim` is maybe nicer.

Also, once we have #17, the type (`int`) would change here.

nn/conformer.py
```diff
@@ -44,18 +43,17 @@ class _ConformerConvBlock(nn.Module):
...
     :param dim_model:
     :param kernel_size:
-    :param l2:
     """
     super().__init__()
 
-    self.positionwise_conv1 = nn.Linear(n_out=dim_model * 2, l2=l2)
-    self.depthwise_conv = nn.Conv(n_out=dim_model, filter_size=(kernel_size,), groups=dim_model, l2=l2, padding='same')
-    self.positionwise_conv2 = nn.Linear(n_out=dim_model, l2=l2)
+    self.positionwise_conv1 = nn.Linear(n_out=dim_model * 2)
```
albertz, line 53: Note, once we have #17, this doesn't really work like this.

mmz33: so better to implement GLU in the way you suggested before by splitting this into 2 modules?

albertz: Not sure. Maybe easier for now. But we probably need to make something like this work in any case at some point. Also, I wonder which variant is actually more efficient...

nn/conformer.py
```diff
@@ -0,0 +1,213 @@
...
+    self.positionwise_conv2 = nn.Linear(n_out=dim_model)
+
+    if batch_norm_other_opts is None:
+      batch_norm_other_opts = {}
+    self.batch_norm = nn.BatchNorm(
+      epsilon=batch_norm_eps, momentum=batch_norm_momentum, update_sample_only_in_training=True,
+      delay_sample_update=True, **batch_norm_other_opts)
+
+  @staticmethod
+  def _glu(v: nn.LayerRef):
```
albertz, line 64: This should be moved over to `math_.py`.

nn/conformer.py
albertz: I think the variant with `dim_model * 2` is simpler though (the code is simpler). So maybe anyway use that. And we just need to make sure that the dim tag logic works here.

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5,
+      batch_norm_momentum: float = 0.1, batch_norm_other_opts=None):
```
albertz, line 46: The indent is inconsistent to the other params. Either use hanging indent here (although I don't like this, but for the function signature it is maybe ok. Or indent all params.
```suggestion
  def __init__(
      self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5,
      batch_norm_momentum: float = 0.1, batch_norm_other_opts=None):
```

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class _ConformerConvSubsampleLayer(nn.Module):
+  """
+  Conv 2D block with optional max-pooling
+  """
+
+  def __init__(self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
+      channel_sizes: List[int], dropout: float = 0.3, activation: str = 'relu',
+      padding: str = 'same'):
```
albertz, line 80: Better indent:
```suggestion
  def __init__(
      self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
      channel_sizes: List[int], dropout: float = 0.3, activation: str = 'relu',
      padding: str = 'same'):
```

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    out = nn.merge_dims(x, axes='static')
+    return out
+
+
+class ConformerEncoderLayer(nn.Module):
+  """
+  Represents a conformer block
+  """
+
+  def __init__(self, conv_kernel_size: int, activation_ff, dim_ff: int, dropout: float, att_dropout: float,
```
albertz, line 116: Indent

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    # last LN layer
+    return nn.layer_norm(x_ffn2_out)
+
+
+class ConformerEncoder(nn.Module):
+  """
+  Represents Conformer encoder architecture
+  """
+
+  def __init__(self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size: int = 32,
```
albertz, line 171: Indent

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    self.dropout = dropout
+    self.pool_sizes = pool_sizes
+
+    self.conv_layers = nn.ModuleList()
+    for filter_size, channel_size in zip(filter_sizes, channel_sizes):
+      self.conv_layers.append(
+        nn.Conv(activation=activation, filter_size=filter_size, n_out=channel_size, padding=padding))
+
+  def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
+    x = nn.split_dims(inp, axis='F', dims=(-1, 1))
```
albertz, line 100: This likely has to change with dim tags #17.

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    :param activation:
+    :param padding:
+    """
+    super().__init__()
+
+    self.dropout = dropout
+    self.pool_sizes = pool_sizes
+
+    self.conv_layers = nn.ModuleList()
+    for filter_size, channel_size in zip(filter_sizes, channel_sizes):
```
albertz, line 95: Maybe you want `assert len(filter_sizes) == len(channel_sizes)` before?

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    x_conv2 = self.positionwise_conv2(x_swish)
+    return x_conv2
+
+
+class _ConformerConvSubsampleLayer(nn.Module):
+  """
+  Conv 2D block with optional max-pooling
+  """
+
+  def __init__(self, filter_sizes: List[Tuple[int, ...]], pool_sizes: Union[List[Tuple[int, ...]], None],
```
albertz, line 78: `None` is not the correct type. But anyway use `Optional[...]` instead.

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+      channel_sizes: List[int], dropout: float = 0.3, activation: str = 'relu',
```
albertz, line 79: Don't use `activation` as a string. And also don't pass it to `Conv` below but instead explicitly apply it afterwards.

nn/math_.py
```diff
@@ -27,6 +27,13 @@ def gelu(x: LayerRef) -> Layer:
   return activation(x, activation="gelu")
 
 
+def glu(v: LayerRef):
```
albertz, line 30:
```suggestion
def glu(v: LayerRef) -> Layer:
```

nn/conformer.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    super().__init__()
+
+    self.dropout = dropout
+    self.pool_sizes = pool_sizes
+
+    self.conv_layers = nn.ModuleList()
+    for filter_size, channel_size in zip(filter_sizes, channel_sizes):
+      self.conv_layers.append(
+        nn.Conv(activation=activation, filter_size=filter_size, n_out=channel_size, padding=padding))
```
albertz, line 97: With #61, this would be `nn.Conv1d`.

nn/conformer.py
albertz: You don't actually expect `Tuple[int, ...]` for the filter sizes and pool sizes, right? You always expect 1D convolution and pooling, so it would be `Tuple[int]` (a single int). But also, with #61 and when you use `nn.Conv1d`, you can simply use a single `int` instead.

nn/conformer.py
albertz: #62

nn/conformer.py
mmz33: Yes I agree.

nn/conformer.py
mmz33: This is 2D Conv so a tuple of ints is expected.

albertz: But then specify it explicitly: `Tuple[int,int]`

And really? This is 2D conv?

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(
+      self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5, batch_norm_momentum: float = 0.1,
+      batch_norm_other_opts=None):
```
albertz, line 47: Sorry, I think to not get a warning here, you need to indent one more, right?
```suggestion
        self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5, batch_norm_momentum: float = 0.1,
        batch_norm_other_opts=None):
```

nn/conformer.py
mmz33: yes. Also similar to ESPNet:
- https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/conformer_encoder.py#L145
- https://github.com/espnet/espnet/blob/a1a04e05a828353c74775327095f78273faba4f3/espnet/nets/pytorch_backend/transformer/subsampling.py#L59

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(
+      self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5, batch_norm_momentum: float = 0.1,
```
albertz, line 46: I don't find my earlier comments on this anymore. But you still have the batch norm options. I think they should be removed (see my earlier comments).

mmz33: you mean to just have `batch_norm_opts` dict?

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+
+class _ConformerConvBlock(nn.Module):
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(
+        self, out_dim: int, kernel_size: int, batch_norm_eps: float = 1e-5, batch_norm_momentum: float = 0.1,
+        batch_norm_other_opts=None):
```
albertz, line 47: See my other comments on batch norm options.

mmz33: ok but the values for `batch_norm_eps` and `batch_norm_momentum` won't be specified in `nn.BatchNorm` right? otherwise we can not pass them in `batch_norm_opts` as this would be duplicate params.

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+
+    self.linear = nn.Linear(n_out=enc_key_dim, with_bias=False)
+
+    self.conformer_blocks = nn.Sequential(
+      encoder_layer(
+        conv_kernel_size=conv_kernel_size, activation_ff=activation_ff, dim_ff=dim_ff, dropout=dropout,
+        att_dropout=att_dropout, enc_key_dim=enc_key_dim, num_heads=num_heads
+      )
+      for _ in range(num_blocks)
+    )
```
albertz, line 208: I prefer:
```suggestion
        att_dropout=att_dropout, enc_key_dim=enc_key_dim, num_heads=num_heads)
      for _ in range(num_blocks))
```

nn/math_.py
```diff
@@ -27,6 +27,13 @@ def gelu(x: LayerRef) -> Layer:
   return activation(x, activation="gelu")
 
 
+def glu(v: LayerRef) -> Layer:
```
albertz, line 30: I think we should also have `axis` here (with `axis="F"` as default).

nn/conformer.py
albertz: Yes.

nn/math_.py
mmz33: btw i am not sure if you checked issue #66. I am getting error when using `nn.split` here.

nn/conformer.py
albertz: What do you mean? They are specified in `nn.BatchNorm`, with reasonable defaults. Where are they duplicated?

nn/conformer.py
```diff
@@ -0,0 +1,215 @@
...
+    x_conv1 = self.positionwise_conv1(inp)
+    x_act = nn.glu(x_conv1)
+    x_depthwise_conv = self.depthwise_conv(x_act)
+    x_bn = self.batch_norm(x_depthwise_conv)
+    x_swish = nn.swish(x_bn)
+    x_conv2 = self.positionwise_conv2(x_swish)
+    return x_conv2
+
+
+class _ConformerConvSubsampleLayer(nn.Module):
```
albertz, line 74: We usually don't use `Layer` in the module name.

albertz: (Except `ConformerEncoderLayer` or `nn.TransformerEncoderLayer` where it is ok.)

nn/conformer.py
mmz33: Yeah nevermind. Should be resolved now.

nn/conformer.py
```diff
@@ -120,7 +122,7 @@ class ConformerEncoderLayer(nn.Module):
 
   def __init__(
         self, conv_kernel_size: int, activation_ff: Callable[[nn.LayerRef], nn.LayerRef], dim_ff: int, dropout: float,
-        att_dropout: float, enc_key_dim: int, num_heads: int):
+        att_dropout: float, enc_key_dim: int, num_heads: int, batch_norm_opts: Optional[Dict[str, Any]]):
```
albertz, line 125: When it is optional, you should set the default `None`.

nn/conformer.py
```diff
@@ -79,9 +79,9 @@ class _ConformerConvSubsample(nn.Module):
   """
 
   def __init__(
-        self, filter_sizes: List[Tuple[int, int]], pool_sizes: Optional[List[Tuple[int, int]]],
-        channel_sizes: List[int], dropout: float = 0.3, activation: Callable[[nn.LayerRef], nn.LayerRef] = nn.relu,
-        padding: str = 'same'):
+        self, filter_sizes: List[Tuple[int, int]], channel_sizes: List[int],
+        pool_sizes: Optional[List[Tuple[int, int]]] = None, dropout: float = 0.3,
```
albertz, line 83: I wonder about the `dropout` default. Is this really so universal? Maybe it should be explicit? Or 0 by default?

albertz:
Note, there is some initial self attention implementation now (#52), although this is also very much untested, and probably will also change with #17.

nn/conformer.py
mmz33: Hmm yes this should be changed. ESPNet uses here the same dropout rate as the dropout layer applied just before the conformer blocks which is 0.1 by default.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class ConformerEncoder(nn.Module):
+  """
+  Represents Conformer encoder architecture
+  """
+
+  def __init__(
+        self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size: int = 32,
+        activation_ff: Callable[[nn.LayerRef], nn.LayerRef] = nn.swish, dim_ff: int = 512, dropout: float = 0.1,
```
albertz, line 182: Here you have also dropout defaults, and also att_dropout.
I'm just curious. Is this from the paper? Or ESPNet? Or both? Are these defaults good?
Also for dimension, num heads.
Does PyTorch Transformer has such defaults?

mmz33: I think this was just a good starting point when I started doing experiments with conformer. But we can follow the parmas of ESPNet best conformer config as here: https://github.com/espnet/espnet/blob/master/egs2/librispeech/asr1/conf/tuning/train_asr_conformer7_n_fft512_hop_length256.yaml

albertz: But I wonder whether we should set any defaults at all. Or which defaults. It seems kind of arbitrary.
Although I can see that some defaults make sense.
Is this the same as in the original paper?

Why is there no default for `encoder_layer` and `num_blocks`?
Btw, instead of `num_blocks`, maybe `num_layers` is better?
(Like [PyTorch Transformer](https://pytorch.org/docs/stable/generated/torch.nn.Transformer.html#torch.nn.Transformer).)

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+    self.positionwise_conv1 = nn.Linear(n_out=out_dim * 2)
+    self.depthwise_conv = nn.Conv(n_out=out_dim, filter_size=(kernel_size,), groups=out_dim, padding='same')
+    self.positionwise_conv2 = nn.Linear(n_out=out_dim)
+
+    batch_norm_opts = copy.deepcopy(batch_norm_opts)
+    if batch_norm_opts is None:
+      batch_norm_opts = {}
+    epsilon = batch_norm_opts.pop('epsilon', 1e-5)
+    momentum = batch_norm_opts.pop('momentum', 0.1)
```
albertz, line 61: I don't understand. Why do you pop it here?

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+    self.dropout = dropout
+
+    self.conv_subsample_layer = _ConformerConvSubsample(
+      filter_sizes=[(3, 3), (3, 3)], pool_sizes=[(2, 2), (2, 2)], channel_sizes=[enc_key_dim, enc_key_dim],
+      dropout=dropout)
+
+    self.linear = nn.Linear(n_out=enc_key_dim, with_bias=False)
+
+    self.conformer_blocks = nn.Sequential(
```
albertz, line 206: Sorry, maybe I suggested it badly earlier, I think we should call it `layers` here instead, to be consistent with `encoder_layer`. Also no need to explicitly put the `conformer_` prefix. The class already is named like that.
```suggestion
    self.layers = nn.Sequential(
```

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    return nn.layer_norm(x_ffn2_out)
+
+
+class ConformerEncoder(nn.Module):
+  """
+  Represents Conformer encoder architecture
+  """
+
+  def __init__(
+        self, encoder_layer: nn.Module, num_blocks: int, conv_kernel_size: int = 32,
```
albertz, line 181: Is `encoder_layer` supposed to be a class (derived from `nn.Module`) or an instance of `nn.Module` (like you specified it here)?

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    :param out_dim:
+    :param kernel_size:
+    """
+    super().__init__()
+
+    self.positionwise_conv1 = nn.Linear(n_out=out_dim * 2)
+    self.depthwise_conv = nn.Conv(n_out=out_dim, filter_size=(kernel_size,), groups=out_dim, padding='same')
+    self.positionwise_conv2 = nn.Linear(n_out=out_dim)
+
+    batch_norm_opts = copy.deepcopy(batch_norm_opts)
```
albertz, line 57: Why is a `deepcopy` needed here?
Why not just:
```suggestion
    batch_norm_opts = dict(batch_norm_opts)
```

nn/conformer.py
albertz: In the Transformer code (#55), it is an instance (so as you specified it here). So I guess this is ok then. But the usage of it is wrong. I will comment below.

nn/conformer.py
```diff
@@ -0,0 +1,217 @@
...
+    self.conv_subsample_layer = _ConformerConvSubsample(
+      filter_sizes=[(3, 3), (3, 3)], pool_sizes=[(2, 2), (2, 2)], channel_sizes=[enc_key_dim, enc_key_dim],
+      dropout=dropout)
+
+    self.linear = nn.Linear(n_out=enc_key_dim, with_bias=False)
+
+    self.conformer_blocks = nn.Sequential(
+      encoder_layer(
+        conv_kernel_size=conv_kernel_size, activation_ff=activation_ff, dim_ff=dim_ff, dropout=dropout,
+        att_dropout=att_dropout, enc_key_dim=enc_key_dim, num_heads=num_heads, batch_norm_opts=batch_norm_opts)
```
albertz, line 209: This is wrong. `encoder_layer` is already an instance, so you cannot pass the options again.
Instead, you would just copy the layer here.
See the corresponding code for Transformers (#55). It should anyway be consistent to that.
So:
```suggestion
      copy.deepcopy(encoder_layer)
```

albertz:
Note, I implemented already some initial variant of self-attention (#52). So we can already partly test this.

Please rebase.

nn/conformer.py
mmz33: > Is this the same as in the original paper?
Not everything. The number of layers is 16 in the original paper. I think the feed forward dimension is not mentioned there. I would suggest to use ESPNet parameters for now. I think i also used the same parameters for the conformer-based hybrid model.

albertz: Ok. But note, as usual, for default values, once we settled on these, we cannot easily change them anymore (unless we introduce sth like behavior version here as well...). So they should be reasonable. Nothing experimental.

mmz33: > Why is there no default for encoder_layer and num_blocks?

Similar to [PyTorch TransformerEncoder](https://pytorch.org/docs/stable/generated/torch.nn.TransformerEncoder.html)

nn/conformer.py
mmz33: Because I need to set the default values. What do you suggest otherwise? If I set them directly in `nn.BatchNorm` then it is not possible to update them via `batch_norm_opts` right?

nn/conformer.py
mmz33: Ok.

nn/conformer.py
mmz33: because I do `pop`

nn/conformer.py
albertz: I don't understand. Why should it not be possible? "Default" means that it is used when you don't specify it, but you can specify it, and then it will use what you specified instead. Like usual.

nn/conformer.py
albertz: `dict(...)` will create a shallow copy. So `pop` would be fine. I just wonder why you do a deep copy.

nn/conformer.py
albertz: But there the code does not create the encoder layer at all, so all the other options are also not needed. Why don't you make it consistent to that anyway?

nn/conformer.py
```diff
@@ -178,40 +179,27 @@ class ConformerEncoder(nn.Module):
...
-    :param ff_dim:
-    :param dropout:
-    :param att_dropout:
-    :param enc_key_dim:
-    :param att_n_heads:
     """
     super().__init__()
 
-    self.dropout = dropout
+    self.dropout = getattr(encoder_layer, 'dropout')
```
mmz33, line 189: I am not sure what is the best way to handle these two attributes. We can pass the dropout as parameter to this class instead so we can set a different dropout rate for the conv subsampling layer. I followed ESPNet recipe for that.

mmz33: So in the Conformer architecture, there is a conv subsampling layer, linear layer, and dropout layer before the conformer blocks. These layers are part of the ConfromerEncoder. The question is if we should decouple this.

albertz: Good question. Definitely you should not use `getattr` here. You might also restrict the type of `encoder_layer` to `ConformerEncoderLayer`. Then you could just do:
```suggestion
    self.dropout = encoder_layer.dropout
```
Anyone who wants to use another custom encoder layer type can simply use his own class. It doesn't make sense that `ConformerEncoder` is so generic here.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+Conformer code.
+Ref: https://arxiv.org/abs/2005.08100
+"""
+
+from typing import Tuple, List, Callable, Optional, Dict, Any, Union
+from .. import nn
+import copy
+
+
+class _PositionwiseFeedForward(nn.Module):
```
albertz, line 11: I wonder whether it make sense to have this public, like:
```suggestion
class ConformerPositionwiseFeedForward(nn.Module):
```

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+
+  def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
+    x_ff1 = self.linear_ff(inp)
+    x_act = self.activation(x_ff1)
+    x_drop = nn.dropout(x_act, dropout=self.dropout)
+    x_ff2 = self.linear_out(x_drop)
+    return x_ff2
+
+
+class _ConformerConvBlock(nn.Module):
```
albertz, line 40: I would make this public:
```suggestion
class ConformerConvBlock(nn.Module):
```

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+  """
+  Conformer convolution block
+      FF -> GLU -> depthwise conv -> BN -> Swish -> FF
+  """
+
+  def __init__(self, out_dim: int, kernel_size: int, batch_norm_opts: Optional[Dict[str, Any]] = None):
+    """
+    :param out_dim:
+    :param kernel_size:
```
albertz, line 49: This is incomplete.
But anyway, when you don't really add a documentation for the params, you can also leave this away.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+    x_conv1 = self.positionwise_conv1(inp)
+    x_act = nn.glu(x_conv1)
+    x_depthwise_conv = self.depthwise_conv(x_act)
+    x_bn = self.batch_norm(x_depthwise_conv)
+    x_swish = nn.swish(x_bn)
+    x_conv2 = self.positionwise_conv2(x_swish)
+    return x_conv2
+
+
+class _ConformerConvSubsample(nn.Module):
```
albertz, line 74: Also public:
```suggestion
class ConformerConvSubsample(nn.Module):
```

Btw, is this the same VGG style subsampling we also use in other cases? (E.g. our standard attention models.) If it is just the same, maybe we should move this out and call it `VGGSubsample` or so.

nn/conformer.py
```diff
@@ -0,0 +1,204 @@
...
+
+  def __init__(self, encoder_layer: Union[ConformerEncoderLayer, Any], num_blocks: int):
+    """
+    :param encoder_layer:
+    :param num_blocks:
+    """
+    super().__init__()
+
+    self.dropout = getattr(encoder_layer, 'dropout')
+    out_dim = getattr(encoder_layer, 'out_dim')
```
albertz, line 189: It should be consistent, also an attrib.
```suggestion
    self.out_dim = encoder_layer.out_dim
```

albertz:
I don't see any positional encoding (relative, absolute, or any) in the code. Is this the standard model without? I think ESPNet uses it.

(Btw, also the original PyTorch Transformer lacks this. There is some interesting relevant on-going discussion here: https://github.com/pytorch/pytorch/issues/24826.)

nn/conformer.py
mmz33: yes this is the same as VGG style subsampling. This is not specific for conformer.

nn/conformer.py
```diff
@@ -37,16 +37,17 @@ def forward(self, inp: nn.LayerRef) -> nn.LayerRef:
...
       FF -> GLU -> depthwise conv -> BN -> Swish -> FF
   """
 
   def __init__(self, out_dim: int, kernel_size: int, batch_norm_opts: Optional[Dict[str, Any]] = None):
     """
-    :param out_dim:
-    :param kernel_size:
+    :param out_dim: output feature dimension
+    :param kernel_size: kernel size of depthwise convolution
+    :param batch_norm_opts: batch norm options
```
albertz, line 50: I really don't like descriptions which just basically repeat what the argument name already says...

nn/conformer.py
```diff
@@ -0,0 +1,205 @@
...
+
+    self.positionwise_conv1 = nn.Linear(n_out=out_dim * 2)
+    self.depthwise_conv = nn.Conv(n_out=out_dim, filter_size=(kernel_size,), groups=out_dim, padding='same')
+    self.positionwise_conv2 = nn.Linear(n_out=out_dim)
+
+    if batch_norm_opts is None:
+      batch_norm_opts = {}
+    batch_norm_opts = batch_norm_opts.copy()
+    batch_norm_opts.setdefault('epsilon', 1e-5)
+    batch_norm_opts.setdefault('momentum', 0.1)
```
albertz, line 62:
```suggestion
```

albertz:
Merged now. Let's further improve this in later PRs or commits.

albertz:
A question on the ConformerConvSubsample: I think this is not exactly described in the original paper, right? So where is it exactly described? Where did you base this on?

mmz33:
Yes, in the original paper it is not described. The one implemented here was based on ESPNet code. See here:
- https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/conformer_encoder.py#L151
- https://github.com/espnet/espnet/blob/f6a2522ad0035451e1edf69b7c92cba272b25871/espnet/nets/pytorch_backend/transformer/subsampling.py#L44

albertz:
But in the original paper, there were some references to other paper they referred to on the preprocessing. I think it is explained in those other papers, or not?

Also, in your implementation here, you use maxpooling as far as I see. But this is different to ESPnet. In ESPnet, they don't use any pooling but just striding instead. Where do you have the pooling from?

albertz:
Also, in ESPnet there are some other variants, including VGG2L (and I think VGG-style was also mentioned in the Conformer paper, or the one it refers to?). VGG2L looks also similar to @christophmluscher 's hybrid baseline?

albertz:
Further, in ESPnet ([`ConformerEncoder`](https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/conformer_encoder.py)), there is pos encoding at the end of the preprocessing block: [here](https://github.com/espnet/espnet/blob/f6a2522ad0035451e1edf69b7c92cba272b25871/espnet/nets/pytorch_backend/transformer/subsampling.py#L66)

This is usually [`RelPositionalEncoding`](https://github.com/espnet/espnet/blob/f6a2522ad0035451e1edf69b7c92cba272b25871/espnet/nets/pytorch_backend/transformer/embedding.py#L259) (e.g. in [train_asr_conformer10_hop_length160](https://github.com/espnet/espnet/blob/f6a2522ad0035451e1edf69b7c92cba272b25871/egs2/librispeech/asr1/conf/tuning/train_asr_conformer10_hop_length160.yaml)).

Then, for the self-attention, it usually uses [`RelPositionMultiHeadedAttention`](https://github.com/espnet/espnet/blob/f6a2522ad0035451e1edf69b7c92cba272b25871/espnet/nets/pytorch_backend/transformer/attention.py#L212).

albertz:
Also related: https://github.com/espnet/espnet/pull/2816

mmz33:
> Where do you have the pooling from?
Ok then it is not exactly the same. you are right they use striding instead.

PR #73, opened 2021-11-10
Updated Tests + Deepcopy test

Atticus1806:
Test for the function of `deepcopy`.


Atticus1806:
So I changed the tests to follow the `nn.` imports now, not sure if this breaks stuff, it does not find LSTM so I guess this is somewhere else now?

albertz (approved):

PR #102, opened 2022-02-06
Implement Cond

### LINKED ISSUE #24: Cond wrapper
Similar as #23, some API with `with Cond(...) as cond_obj:`, which corresponds to `if ...:`, and then some further `with cond_obj.false_branch():`, which corresponds to the `else:` branch.

Example:
```
x = ... # whatever
cond = ...  # scalar tensor, True or False
with Cond(cond) as cond_obj:
  y = mod_true_case(x)
with cond_obj.false_branch():
  y = mod_false_case(x)
```

But this is not so clear yet.

albertz:
Fix #24

### DIFF
--- a/nn/cond.py
+++ b/nn/cond.py
@@ -17,10 +17,20 @@ class Cond:
   Example::
 
       with Cond(cond) as cond_obj:
+        cond_obj.true = mod_true_case(x)
+        cond_obj.false = mod_false_case(x)
+        y = cond_obj.result
+
+  Corresponds to::
+
+      if cond:
         y = mod_true_case(x)
-        cond_obj.else(y)
+      else:
         y = mod_false_case(x)
-        y = cond_obj.end(y)
+
+  The context scope has two states corresponding to the True and False computation branch.
+  The initial state is the True branch.
+  Assigning ``cond_obj.true`` has the side effect of switching the computation to the False branch.
   """
 
   def __init__(self, condition: nn.Tensor, *, name: str = "cond"):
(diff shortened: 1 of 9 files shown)


nn/cond.py
```diff
@@ -0,0 +1,155 @@
...
+  Only one branch will be executed, and the condition needs to be a bool scalar.
+  This wraps to :class:`CondLayer` in RETURNN and to ``tf.cond`` in TensorFlow.
+
+  Example::
+
+      with Cond(cond) as cond_obj:
+        y = mod_true_case(x)
+        cond_obj.else(y)
+        y = mod_false_case(x)
+        y = cond_obj.end(y)
```
Zettelkasten, line 23: This is outdated, and not what this PR actually implements, right?

albertz: Yes, I'm going to update this. Actually I was still not really sure which API to implement and just wanted to test it as part of the test cases. Some suggestion have been made in #24. Although maybe I stick to the current implementation now. I think it's fine.

tests/test_nn_cond.py
```diff
@@ -0,0 +1,78 @@
...
+    @nn.scoped
+    def __call__(self, x: nn.Tensor) -> nn.Tensor:
+      with nn.Cond(nn.length(x, axis=nn.batch_dim) % 2 == 0) as cond:
+        cond.true = self.linear(x)
+        cond.false = self.linear(x * 2.)
+        x = cond.result
+      return x
+
+  config, net_dict = dummy_config_net_dict(net=_Net())
+  dummy_run_net(config)
```
albertz, line 52: The test also should check the right variable name scope.

PR #108, opened 2022-02-11
Fix random Parameter init outside name ctx

### LINKED ISSUE #109: Parameter random init outside name ctx does not work
This happens e.g. for
```
model = nn.Linear(out_dim, in_dim=in_dim)
```
outside a `NameCtx`.

Or maybe we want to disallow this? I.e. simply require that the model (all params) are always defined inside a `NameCtx`.

Or lazily assign the parent name ctx and the name itself later, similar as we do for `Parameter`?

PR in #108.

albertz:
Fix #109.

Fix non-lazy init of Linear.

Currently only the corresponding test case.

Or maybe we want to disallow this? Or lazily assign the parent name ctx land the name itself later, similar as we do for `Parameter`?

### DIFF
--- a/tests/test_nn_base.py
+++ b/tests/test_nn_base.py
@@ -36,6 +36,28 @@ def test_simple_net_linear():
   dummy_run_net(config)
 
 
+def test_simple_net_linear_non_lazy():
+  class _Net(nn.Module):
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


albertz:
Obsolete via #215.

PR #127, opened 2022-04-08
Update gammatone net dict

vieting:
Some updates to allow setting parameters when creating the Gammatone net dict.

The default network is changed by this. I'm not sure if anyone is using this, if yes should we use `BehaviorVersion` here?

The changes are for default behavior are
- removing `'gammatone_filterbank_padding'` which doesn't belong here from my point of view (this was taken from one of my configs but it's too specific here)
- using dim tags for split and merge
- setting `'in_spatial_dims'` and `'out_dim'` for convolutions
- adding `"from"` where it was missing

I could also change `'` -> `"` because that's used mostly in RETURNN, but didn't want to make the diff too complicated initially.

### DIFF
--- a/asr/gt.py
+++ b/asr/gt.py
@@ -9,6 +9,7 @@ Code by Peter Vieting, and adopted.
 import tensorflow as tf
 import numpy
 from returnn.tf.network import TFNetwork, ExternData
+from returnn.tf.util.data import FeatureDim
 
 tf1 = tf.compat.v1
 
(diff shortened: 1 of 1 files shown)


albertz:
Oh, you are working on this? I don't think anyone is really using it. But it would be nice if this is in a good form to be used. Is it?

Yes, this should work with a recent behavior version.

You set this PR as a draft. So it's not ready yet? Don't forget to mark it ready.

vieting:
Yes, I use it in some of my setups but in a local version only so far. It would be nicer to have it in a central place and since something similar already exists here, I thought we could fix this.

I set it as a draft because I wasn't sure if we should check the behavior version here and only apply my changes for a recent behavior version or if we can just change the default behavior as this hasn't been used so far anyway.

albertz:
> I set it as a draft because I wasn't sure if we should check the behavior version here and only apply my changes for a recent behavior version 

I don't exactly understand. This is just about the net dict? There is no behavior version here. What exactly do you want to check?

You should (of course) make the net dict such that it works correctly with the latest behavior version.

> or if we can just change the default behavior as this hasn't been used so far anyway.

As no-one is using it, I think this it's fine to change behavior.

Ah, now I think I get what you mean by checking behavior version. This does not really make sense as there is no behavior version in general. This code is not necessarily (even maybe commonly) not even executed inside RETURNN but maybe by Sisyphus to create a net dict or so.

So far, we did not need to deal with such issues (the motivation behind behavior version in RETURNN) here in returnn_common at all because it has not come up yet. If we ever need it (e.g. fix bad default behavior), I think we need to introduce a new separate behavior version for returnn_common in some way. But let's discuss this at some later point when we really need it.

vieting:
Ok, then this is ready from my side.

albertz (approved):

PR #130, opened 2022-04-13
Followup README changes

Atticus1806:
When doing my mini example, I realized that the example is missing some stuff:

No `super().__init__()` call, which caused my tests to fail with `MyModuleBlock has no attribute "calls"`
No dropout axis. Here I only chose just `nn.batch_dim` because there is no `time_dim` in the example.
`self.linear_out` was called with `x` instead of `y` which made the part beforehand irrelevant. I think this was just a typo and not inteded like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ class MyModelBlock(nn.Module):
     y = self.linear_hidden(y)
     y = nn.sigmoid(y)
     y = self.linear_out(y)
-    y = nn.dropout(y, dropout=self.dropout, axis=[nn.batch_dim])
+    y = nn.dropout(y, dropout=self.dropout, axis=nn.any_feature_dim)
     return x + y
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,15 +35,16 @@ class MyModelBlock(nn.Module):
     y = self.layer_norm(x)
     y = self.linear_hidden(y)
     y = nn.sigmoid(y)
-    y = self.linear_out(x)
-    y = nn.dropout(y, dropout=self.dropout)
+    y = self.linear_out(y)
+    y = nn.dropout(y, dropout=self.dropout, axis=[nn.batch_dim])
```
albertz, line 39: This doesn't really make sense. You don't want to dropout whole sequences. You usually would want this:
```suggestion
    y = nn.dropout(y, dropout=self.dropout, axis=nn.any_feature_dim)
```

albertz (approved):

PR #146, opened 2022-05-02
Typing of combine layer

Atticus1806:
I think the combine layer typing should have `Sequence[nn.Tensor]` instead of `nn.Tensor` typing (just like the `_combine`)

### DIFF
--- a/nn/math_.py
+++ b/nn/math_.py
@@ -195,7 +195,7 @@ def gating(x: nn.Tensor, *, axis: Optional[nn.Dim] = None,
 
 # noinspection PyShadowingBuiltins,PyShadowingNames
 def combine(
-             *sources: Union[nn.Tensor, nn.RawTensorTypes],
+             *sources: Union[Sequence[nn.Tensor], nn.RawTensorTypes],
              kind: str,
              allow_broadcast_all_sources: Union[bool, nn.NotSpecified] = nn.NotSpecified,
              name: Optional[Union[str, nn.NameCtx]] = None) -> nn.Tensor:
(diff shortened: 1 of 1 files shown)


Atticus1806:
I am not sure about the raw tensor types but I guess I assume its teh same there

nn/math_.py
```diff
@@ -195,7 +195,7 @@ def gating(x: nn.Tensor, *, axis: Optional[nn.Dim] = None,
 
 # noinspection PyShadowingBuiltins,PyShadowingNames
 def combine(
-             *sources: Union[nn.Tensor, nn.RawTensorTypes],
+             *sources: Union[Sequence[nn.Tensor], Union[nn.RawTensorTypes]],
```
albertz, line 198: First of all, I assume you mean this?
```suggestion
             *sources: Sequence[Union[nn.Tensor, nn.RawTensorTypes]],
```
But then, I think this is not how it works. Did you actually check (e.g. in PyCharm)? As far as I know, and PyCharm also handles it this way, the type refers to each element of `sources` but not `sources` itself. So it is already correct.

Atticus1806: Yes I did. Without the change Pycharm marked `nn.combine([x, mu], kind="sub", allow_broadcast_all_sources=True)` where `x` and `mu` are tensors as wrong type.

```Expected type 'Union[Tensor, int, float, complex, number, ndarray, bool, str]', got 'List[Tensor]' instead```

albertz: But `nn.combine([x, mu], kind="sub")` is wrong, and PyCharm gives the correct warning?

It is `nn.combine(x, mu, kind="sub")`.

With your change here, you would allow sth like `nn.combine([a, b], [c, d], kind="sub")` which does not really make sense.

Atticus1806: You are right. I did not see the change made to the input to the layer compared to `_combine`, where it worked like this, was assuming it stayed the same for now.

Issue #142, opened 2022-04-29
nn.normalize missing


albertz:
Corresponding to `NormLayer` (but maybe simply reimplemented).

(Do not confuse with norm, which would correspond to `MathNormLayer`, which is a separate thing.)

albertz:
Implemented via #153.

PR #153, opened 2022-05-09
Normalize

Atticus1806:
Implementation for #142 

Pretty much just replicating the math part of the `NormLayer` from returnn.

I think we should not do scale and bias here (can be done explicitly be the user afterwards).

I am unsure about the `keepdims=True` of the reduces in returnn, is there similar handling required here too? I dont think `reduce` offers this option, but I am not sure, do we even need this?

### DIFF
--- a/nn/math_.py
+++ b/nn/math_.py
@@ -317,9 +317,9 @@ def cumsum(
   return layer
 
 
-def layer_norm(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
+def normalize(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
   """
-  Calculates layer norm for given layer, based on the input dims
+  Calculates normalization for given layer, based on the input dims
   :param a: input
   :param axes: axes over which the mean and variance are computed
   :param epsilon: epsilon for numerical stability
--- a/nn/math_.py
+++ b/nn/math_.py
@@ -317,9 +317,10 @@ def cumsum(
   return layer
 
 
-def normalize(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
+def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
   """
   Calculates normalization for given layer, based on the input dims
+  
   :param a: input
   :param axes: axes over which the mean and variance are computed
   :param epsilon: epsilon for numerical stability
--- a/nn/math_.py
+++ b/nn/math_.py
@@ -315,3 +315,16 @@ def cumsum(
     name=name)
   del state
   return layer
+
+
+def layer_norm(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
+  """
+  Calculates layer norm for given layer, based on the input dims
+  :param a: input
+  :param axes: axes over which the mean and variance are computed
+  :param epsilon: epsilon for numerical stability
+  """
+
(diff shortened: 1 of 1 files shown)


nn/math_.py
```diff
@@ -315,3 +315,16 @@ def cumsum(
     name=name)
   del state
   return layer
+
+
+def normalize(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
```
albertz, line 320: Our convention is to use singular (`axis`), if both a single or multiple axes are possible. This is because in practice, using a singular axis is more common.

nn/math_.py
```diff
@@ -315,3 +315,16 @@ def cumsum(
...
+  """
+  Calculates normalization for given layer, based on the input dims
+  :param a: input
+  :param axes: axes over which the mean and variance are computed
+  :param epsilon: epsilon for numerical stability
+  """
+
+  mean = nn.reduce(a, axis=axes, mode='mean')
+  variance = nn.reduce((a - mean) ** 2, mode='mean', axis=axes, name='variance')
```
albertz, line 329: Maybe `squared_difference(a, mean)`? Could be slightly faster.

albertz, line 329: Actually, just use:
```suggestion
  mean, variance = nn.moments(a, axis=axis)
```

nn/math_.py
albertz: Also return type missing.
```suggestion
def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6) -> nn.Tensor:
```

nn/math_.py
```diff
@@ -315,3 +315,16 @@ def cumsum(
     name=name)
   del state
   return layer
+
+
+def normalize(a: nn.Tensor, *, axes: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
+  """
+  Calculates normalization for given layer, based on the input dims
```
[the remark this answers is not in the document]
Atticus1806: you mean between params and the info what the function does right? added it now.

nn/math_.py
```diff
@@ -315,3 +315,16 @@ def cumsum(
     name=name)
   del state
   return layer
+
+
+def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6):
```
albertz, line 320: Return type missing.

albertz:
```suggestion
def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6) -> nn.Tensor:
```

nn/math_.py
```diff
@@ -315,3 +315,16 @@ def cumsum(
...
+  """
+  Calculates normalization for given layer, based on the input dims
+  
+  :param a: input
+  :param axes: axes over which the mean and variance are computed
+  :param epsilon: epsilon for numerical stability
```
albertz, line 326:
```suggestion
  :param epsilon: epsilon for numerical stability
  :return: (a - mean) / sqrt(variance + epsilon)
```

nn/math_.py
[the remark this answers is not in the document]
albertz: Exactly here where I put this comment, yes.
Btw, this is always standard Python docstring style. There always should be an empty line between the description and the params.

albertz (approved):

albertz:
Please check the code style warnings.

Atticus1806:
I just saw that we also have a `normalization` module. I put this in `math_` because for me it seemed natural and I didnt know there was a module, but maybe its a better fit in `normalize`?

albertz:
Right. Feel free to move it.

Atticus1806:
cant merge myself, created a pr.

PR #154, opened 2022-05-09
add ref_extern_data_dims_via_global_config flag

JackTemaki:
I changed `self._base_extern_data_dim_refs` from `set` to `list` because otherwise I would get undefined behavior. Maybe I overlooked something but this is how it worked for me now.

### DIFF
--- a/nn/naming.py
+++ b/nn/naming.py
@@ -718,8 +718,9 @@ class ReturnnConfigSerializer:
       code_lines.append(import_str + "\n")
 
     if ref_extern_data_dims_via_global_config:
-      code_lines += ["from returnn.config import get_global_config\n",
-                     "config = get_global_config()\n"]
+      code_lines += [
+        "from returnn.config import get_global_config\n",
+        "config = get_global_config()\n"]
       for value in self._base_extern_data_dim_refs:
         code_lines += [f"{value.py_id_name()} = config.typed_dict[{value.py_id_name()!r}]\n"]
 
--- a/nn/naming.py
+++ b/nn/naming.py
@@ -658,7 +658,7 @@ class ReturnnConfigSerializer:
     self.name_ctx = name_ctx
     self._behavior_version = nn.min_returnn_behavior_version
     self._dim_tags_proxy = ReturnnDimTagsProxy()
-    self._base_extern_data_dim_refs = None  # type: Optional[Set[ReturnnDimTagsProxy.DimRefProxy]]
+    self._base_extern_data_dim_refs = None  # type: Optional[List[ReturnnDimTagsProxy.DimRefProxy]]
 
   def get_complete_py_code_str(self, root_module: nn.Module):
     """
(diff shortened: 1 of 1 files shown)


albertz:
> I changed `self._base_extern_data_dim_refs` from `set` to `list` because otherwise I would get undefined behavior. Maybe I overlooked something but this is how it worked for me now.

I don't understand. Why? This is wrong. It should be a `set`. If there is some other bug somewhere, and this is a workaround (which I don't understand), then: We should not add workarounds but fix the real bugs instead. Or have a very good explanation why this is needed. Not just "it worked for me now".

JackTemaki:
> Or have a very good explanation why this is needed.

A set has no defined order. Sisyphus needs a defined order as otherwise the serialized string always changes, and with it the hashes. Thus I changed this to list. Otherwise we need to add some explicit hash handling.

albertz:
> > Or have a very good explanation why this is needed.
> 
> A set has no defined order. Sisyphus needs a defined order as otherwise the serialized string always changes, and with it the hashes. Thus I changed this to list. Otherwise we need to add some explicit hash handling.

I don't understand. `_base_extern_data_dim_refs` is never serialized.

JackTemaki:
> > > Or have a very good explanation why this is needed.
> > 
> > 
> > A set has no defined order. Sisyphus needs a defined order as otherwise the serialized string always changes, and with it the hashes. Thus I changed this to list. Otherwise we need to add some explicit hash handling.
> 
> I don't understand. `_base_extern_data_dim_refs` is never serialized.

What? This is exactly what this PR is about. Serializing the `_base_extern_data_dim_refs` in the specific network config by writing code that imports it from the base config.

albertz:
> What? This is exactly what this PR is about. Serializing the `_base_extern_data_dim_refs` in the specific network config by writing code that imports it from the base config.

Ah sorry, I just read your description here and thought you just changed `set` to `list`.

JackTemaki:
> > What? This is exactly what this PR is about. Serializing the `_base_extern_data_dim_refs` in the specific network config by writing code that imports it from the base config.
> 
> Ah sorry, I just read your description here and thought you just changed `set` to `list`.

Ah no, this was just meant as additional information.

nn/naming.py
```diff
@@ -705,12 +710,19 @@ def get_ext_net_dict_py_code_str(self, root_module: nn.Module, *, with_imports:
...
     code_lines = []
+
     if with_imports:
       code_lines.append(self._ImportPyCodeStr + "\n")
     for import_str in imports:
       code_lines.append(import_str + "\n")
+
+    if ref_extern_data_dims_via_global_config:
+      code_lines += ["from returnn.config import get_global_config\n",
+                     "config = get_global_config()\n"]
```
albertz, line 722: Do not use hanging indents (same rule as in RETURNN).
```suggestion
      code_lines += [
        "from returnn.config import get_global_config\n",
        "config = get_global_config()\n"]
```

nn/naming.py
```diff
@@ -705,12 +710,19 @@ def get_ext_net_dict_py_code_str(self, root_module: nn.Module, *, with_imports:
...
+      for value in self._base_extern_data_dim_refs:
+        code_lines += [f"{value.py_id_name()} = config.typed_dict[{value.py_id_name()!r}]\n"]
```
albertz, line 724:
```suggestion
        code_lines.append(f"{value.py_id_name()} = config.typed_dict[{value.py_id_name()!r}]\n")
```

albertz (approved):

PR #157, opened 2022-05-10
Add Rint as activation function

Atticus1806:
Adds "tf.math.rint" for rounding as activation function.

### DIFF
--- a/nn/math_.py
+++ b/nn/math_.py
@@ -41,6 +41,11 @@ def floor(x: nn.Tensor) -> nn.Tensor:
   return _activation(x, activation="floor")
 
 
+def rint(x: nn.Tensor) -> nn.Tensor:
+  """round"""
+  return _activation(x, activation="rint")
+
+
 def relu(x: nn.Tensor) -> nn.Tensor:
   """ReLU"""
   return _activation(x, activation="relu")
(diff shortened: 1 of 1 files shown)


nn/math_.py
```diff
@@ -41,6 +41,11 @@ def floor(x: nn.Tensor) -> nn.Tensor:
   return _activation(x, activation="floor")
 
 
+def rint(x: nn.Tensor) -> nn.Tensor:
+  """round"""
```
albertz, line 45: The docstring is misleading. rint is not round, although very similar. round is another separate function.

Atticus1806: oh true, had it on round first. Will change

Atticus1806:
So I think `rint` came from `numpy` which also calls it `rint`, but at the same time someone implemented `round` in `tf`. From my understanding, in general `round` will always do "bankers rounding", while `rint` current system rounding mode, which defaults to the same.
For me it does not really matter, decided for `rint` because its closer to `numpy` and also does not shadow `round`.

In general (not the way tensorflow implements) `round` rounds to given number of decimals, while `rint` is always to next int.

Atticus1806:
I think Pytorch only has `round`

PR #160, opened 2022-05-16
Remove nn.scoped

### LINKED ISSUE #159: Remove `nn.scoped`
I think we don't really need it, and it has some disadvantages:

- Adds some small boilerplate code.
- Maybe not intuitive to understand.
- Adds to the stack trace for each func call, which makes the stack trace harder to understand.
- Breaks PyCharm auto-completion for such functions.

The reason why we have it is just to get nicer automatic sub network scopes.

I think we can still get the same by automatically inspecting the stack trace and finding all method calls of a module. This is pretty straightforward in Python.

albertz:
Fix #159.

Removes `scoped`.

Also cleans up some related code.

In `NameCtx.current_ctx`, we will derive the scope.

### DIFF
--- a/asr/gt.py
+++ b/asr/gt.py
@@ -56,7 +56,6 @@ class GammatoneV2(nn.Module):
       .reshape((temporal_integration_size, 1, 1))
       .astype(numpy.float32))
 
-  @nn.scoped
   def __call__(
         self,
         raw_samples: nn.Tensor,
--- a/asr/specaugment.py
+++ b/asr/specaugment.py
@@ -7,7 +7,6 @@ from typing import Union, Collection
 from .. import nn
 
 
-@nn.scoped
 def specaugment_v2(x: nn.Tensor, *,
                    spatial_dim: nn.Dim,
                    feature_dim: nn.Dim = nn.NotSpecified,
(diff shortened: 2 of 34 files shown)


albertz:
This PR got too big...

Taking out some semi-unrelated chunks:
- #164 
- #165 
- #166 
- #167 
-

albertz:
Current state:

- SpecAugment tests are failing. Sth looks messed up with the name scope there. I need to investigate. Maybe produce a simpler test case.
- Before this is ready, we should check a bit more if the name spaces are good in most common cases. E.g. tests like `test_specaugment_v2` will get `test_specaugment_v2` as a name scope prefix, which is maybe not nice.

albertz:
- [ ] SpecAugment tests are passing now.
- [ ] Should move `test_specaugment_v2_name_scope_simplify`, maybe to loop tests.
- [ ] Before this is ready, we should check a bit more if the name spaces are good in most common cases. E.g. tests like `test_specaugment_v2` will get `test_specaugment_v2` as a name scope prefix, which is maybe not nice. Also check local funs, I have seen `test_func.<locals>.local_func`.
- [ ] Other cleanups?

Issue #188, opened 2022-06-22
nn.normalize without scale and bias


Atticus1806:
I just noticed that the current implementation of `nn.normalize` does not have the option to add scale and bias. 
In general this of course can be added by the user by
```python
x = nn.normalize(x, axis=x.feature_dim)
bias = nn.Parameter(list(x.shape), trainable=True)
scale = nn.Parameter(list(x.shape), trainable=True)
x = scale * x + bias
```

Still this is different from how its implemented in Returnn, so I wanted to bring this up for discussion whether it makes sense to either document this in `nn.normalize` or actually implement it.

albertz:
I guess we can consider this as closed via #190.

PR #190, opened 2022-06-22
nn.Normalize

Atticus1806:
#188 
adds `Normalize` and documents `normalize`

Note that 
```python
    if isinstance(axis, nn.Dim):
      axis = [axis]
``` 
is because `Parameter` does not support `nn.Dim`, but only `Sequence[nn.Dim]`, maybe we can also just catch this in `Parameter`.

### DIFF
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -188,7 +188,7 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
 
 class Normalize(nn.Module):
   """
-  Wrapper around normalize with additional scale and bias
+  :func:`normalize` with additional scale and bias
   """
 
   def __init__(self,
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -220,7 +220,7 @@ class Normalize(nn.Module):
     norm = normalize(a, axis=self.axis, epsilon=self.epsilon)
     if self.scale is not None:
       norm = self.scale * norm
-    if self.bias:
+    if self.bias is not None:
       norm = norm + self.bias
 
     return norm
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -192,11 +192,10 @@ class Normalize(nn.Module):
   """
 
   def __init__(self, *,
-               param_shape: Sequence[nn.Dim],
+               param_shape: Union[nn.Dim, Sequence[nn.Dim]],
                epsilon: float = 1e-6,
                scale: bool = True, bias: bool = True):
     """
-
     :param param_shape: shape of the scale and bias parameters
     :param epsilon: epsilon for numerical stability
     :param scale: whether to include a trainable scale
@@ -204,15 +203,15 @@ class Normalize(nn.Module):
     """
     super(Normalize, self).__init__()
     self.epsilon = epsilon
+    if isinstance(param_shape, nn.Dim):
+      param_shape = [param_shape]
     self.scale = nn.Parameter(shape=param_shape) if scale else None
     self.bias = nn.Parameter(shape=param_shape) if bias else None
 
   def __call__(self, a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]]):
-
     norm = normalize(a, axis=axis, epsilon=self.epsilon)
     if self.scale is not None:
       norm = self.scale * norm
     if self.bias is not None:
       norm = norm + self.bias
-
     return norm
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -224,4 +224,3 @@ class Normalize(nn.Module):
       norm = norm + self.bias
 
     return norm
-
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -174,6 +174,7 @@ class BatchNorm(nn.Module):
 def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: float = 1e-6) -> nn.Tensor:
   """
   Calculates normalization for given layer, based on the input dims
+  This does not support parameters as in :class:`NormLayer`, cf. :class:`Normalize`
 
   :param a: input
   :param axis: axis over which the mean and variance are computed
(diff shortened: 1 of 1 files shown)


nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+    if isinstance(axis, nn.Dim):
+      axis = [axis]
+    self.scale = nn.Parameter(shape=axis, trainable=True) if scale else None
+    self.bias = nn.Parameter(shape=axis, trainable=True) if bias else None
+
+  def __call__(self, a: nn.Tensor):
+    """
+
+    :param a: input
+    :return:
```
albertz, line 217: Such docstring does not really add any information. Just leave it away. Shorter code can be read faster and thus is easier to understand.

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+    """
+
+    norm = normalize(a, axis=self.axis, epsilon=self.epsilon)
```
albertz, line 220: I don't think it makes sense to use `self.axis` here.

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
 
   mean, variance = nn.moments(a, axis=axis)
   return (a - mean) * nn.rsqrt(variance + epsilon)
+
+
+class Normalize(nn.Module):
+  """
+  Wrapper around normalize with additional scale and bias
```
albertz, line 191: You should refer to the function correctly.
```suggestion
  Wrapper around :func:`normalize` with additional scale and bias
```

albertz: But also, I think describing this as "wrapper" is confusing.
Just say:
```suggestion
  :func:`normalize` with additional scale and bias
```

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+    :param epsilon: epsilon for numerical stability
+    :param scale: whether to include a trainable scale
+    :param bias: whether to include a trainable bias
+    """
+    super(Normalize, self).__init__()
+    self.epsilon = epsilon
+    self.axis = axis
+    if isinstance(axis, nn.Dim):
+      axis = [axis]
+    self.scale = nn.Parameter(shape=axis, trainable=True) if scale else None
```
albertz, line 210: You don't need the `trainable` argument. The code should be consistent to other code.

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+
+  def __call__(self, a: nn.Tensor):
+    """
+
+    :param a: input
+    :return:
+    """
+
+    norm = normalize(a, axis=self.axis, epsilon=self.epsilon)
+    if self.scale:
```
albertz, line 221: I think you should better check for `is not None`:
```suggestion
    if self.scale is not None:
```

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+      norm = self.scale * norm
+    if self.bias:
```
albertz, line 223:
```suggestion
    if self.bias is not None:
```

nn/normalization.py
```diff
@@ -183,3 +184,44 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+      norm = norm + self.bias
+
+    return norm
+
```
albertz, line 227: See failing test.
```
returnn_common/nn/normalization.py:227: WEAK WARNING Pep8CodeStyle: W391: blank line at end of file
```
```suggestion
```

nn/normalization.py
Atticus1806: Why do I not need it? Its default is not `True`.

albertz: It is (on RETURNN side). And in any case, it should be consistent to other code.

Atticus1806: Ah okay, I did not know that.

nn/normalization.py
```diff
@@ -183,3 +184,35 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+    :param scale: whether to include a trainable scale
+    :param bias: whether to include a trainable bias
+    """
+    super(Normalize, self).__init__()
+    self.epsilon = epsilon
+    self.scale = nn.Parameter(shape=param_shape) if scale else None
+    self.bias = nn.Parameter(shape=param_shape) if bias else None
+
+  def __call__(self, a: nn.Tensor, axis: Union[nn.Dim, Sequence[nn.Dim]]):
+
```
albertz, line 211: This empty line is inconsistent to our other code.

nn/normalization.py
```diff
@@ -183,3 +184,35 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+    norm = normalize(a, axis=axis, epsilon=self.epsilon)
+    if self.scale is not None:
+      norm = self.scale * norm
+    if self.bias is not None:
+      norm = norm + self.bias
+
```
albertz, line 217: I don't really see how this empty line adds readability.

nn/normalization.py
```diff
@@ -183,3 +184,35 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
+  """
+  :func:`normalize` with additional scale and bias
+  """
+
+  def __init__(self,
+               param_shape: Sequence[nn.Dim],
+               epsilon: float = 1e-6,
+               scale: bool = True, bias: bool = True):
+    """
+
```
albertz, line 199: There should be no empty line here (unless you have some initial description, which you do not have here).

nn/normalization.py
```diff
@@ -183,3 +184,35 @@ def normalize(a: nn.Tensor, *, axis: Union[nn.Dim, Sequence[nn.Dim]], epsilon: f
...
   mean, variance = nn.moments(a, axis=axis)
   return (a - mean) * nn.rsqrt(variance + epsilon)
+
+
+class Normalize(nn.Module):
+  """
+  :func:`normalize` with additional scale and bias
+  """
+
+  def __init__(self,
```
albertz, line 194: I think all the args should be named args.
```suggestion
  def __init__(self, *,
```

albertz (approved):

Issue #189, opened 2022-06-22
nn.Parameter in list


Atticus1806:
I am currently trying to create an (arbitrary) number of parameters in an `nn.Module`. I would store them in a list since `nn.ModuleList` is not intended for `nn.Parameters`, at least thats what the typing suggests. But when using parameters in a list the parent assignment fails. Example:

```python
def test_parameter_usage():
  nn.reset_default_root_name_ctx()
  time = nn.SpatialDim("time")
  feature = nn.FeatureDim("feature", 1)
  data = nn.get_extern_data(nn.Data("data", dim_tags=[nn.batch_dim, time, feature], dtype="float32"))

  class _Net(nn.Module):

    def __init__(self, feat_dim):
      super(_Net, self).__init__()
      self.scales = [nn.Parameter(shape=[feat_dim], trainable=True)]
      self.biases = [nn.Parameter(shape=[feat_dim], trainable=True)]

    def __call__(self, x):
      return self.scales[0] * x + self.biases[0]

  net = _Net(feature)
  x = net(data)
  x.mark_as_default_output()

  config = nn.get_returnn_config().get_complete_py_code_str(net)
```

fails with: 

```
  File "/u/hilmes/experiments/tts_new_sis/recipe/i6_private/users/hilmes/old/tests.py", line 150, in <module>
    line: test_parameter_usage()
    locals:
      test_parameter_usage = <local> <function test_parameter_usage at 0x7ff493430670>
  File "/u/hilmes/experiments/tts_new_sis/recipe/i6_private/users/hilmes/old/tests.py", line 146, in test_parameter_usage
    line: config = nn.get_returnn_config().get_complete_py_code_str(net)
    locals:
      config = <not found>
      nn = <global> <module 'returnn_common.nn' from '/u/hilmes/experiments/tts_new_sis/recipe/returnn_common/nn/__init__.py'>
      nn.get_returnn_config = <global> <function get_returnn_config at 0x7ff452bfef70>
      get_complete_py_code_str = <not found>
      net = <local> <_Net>
  File "/u/hilmes/experiments/tts_new_sis/recipe/returnn_common/nn/naming.py", line 763, in ReturnnConfigSerializer.get_complete_py_code_str
    line: return (
            self.get_base_extern_data_py_code_str() +
            self.get_ext_net_dict_py_code_str(
              root_module=root_module, with_imports=False, ref_extern_data_dims_via_global_config=False))
    locals:
[...]

PR #191, opened 2022-06-23
ParamList

Atticus1806:
#189 
Similar to `nn.ModuleList`, I could not really think of other functionalities that are required here.


Atticus1806:
Maybe we could also add a test here to check whether accessing the Parameters works. 

Regarding the error message I am not sure how to handle that and would leave that for now.

Atticus1806:
I dont there are reasonable asserts here, so I just left them out.

PR #192, opened 2022-06-27
Fix Scale Init Normalize

Atticus1806:
Scale parameter of `nn.Normalize` was initialized with 0 instead of (as in returnn) with 1.

### DIFF
--- a/nn/normalization.py
+++ b/nn/normalization.py
@@ -205,7 +205,9 @@ class Normalize(nn.Module):
     self.epsilon = epsilon
     if isinstance(param_shape, nn.Dim):
       param_shape = [param_shape]
-    self.scale = nn.Parameter(shape=param_shape) if scale else None
+    if scale is not None:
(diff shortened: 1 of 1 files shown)


nn/normalization.py
```diff
@@ -205,7 +205,9 @@ def __init__(self, *,
     self.epsilon = epsilon
     if isinstance(param_shape, nn.Dim):
       param_shape = [param_shape]
-    self.scale = nn.Parameter(shape=param_shape) if scale else None
+    if scale is not None:
+      self.scale = nn.Parameter(shape=param_shape)
+      self.scale.initial = 1.0
```
albertz, line 210: Multiple serious problems.
* `scale` is bool
* You forgot to set `self.scale` in the else case.

```suggestion
    self.scale = None
    if scale:
      self.scale = nn.Parameter(shape=param_shape)
      self.scale.initial = 1.0
```

albertz (approved):

PR #194, opened 2022-07-19
Reinterpret Set Size Base

Atticus1806:
Similar to `reinterpret_set_sparse_dim` just with the `size_base` parameter instead.

### DIFF
--- a/nn/utils.py
+++ b/nn/utils.py
@@ -160,6 +160,15 @@ def reinterpret_set_sparse_dim(source: nn.Tensor, out_dim: nn.Dim, *, name: str 
     name=name)
 
 
+def reinterpret_set_size_base(source: nn.Tensor, size_base: nn.Tensor, *, name: str = "set_size_base") -> nn.Tensor:
+  """
+  :return: source with sparse_dim set to out_dim
+  """
+  return nn.make_layer(
+    {"class": "reinterpret_data", "size_base": size_base, "from": source},
+    name=name)
+
+
 def check_in_feature_dim_lazy_init(
       source: nn.Tensor, in_dim: Optional[nn.Dim], mod_in_dim: Optional[nn.Dim],
       lazy_init: Callable[[nn.Dim], Any]) -> nn.Tensor:
--- a/nn/utils.py
+++ b/nn/utils.py
@@ -162,7 +162,7 @@ def reinterpret_set_sparse_dim(source: nn.Tensor, out_dim: nn.Dim, *, name: str 
 
 def reinterpret_set_size_base(source: nn.Tensor, size_base: nn.Tensor, *, name: str = "set_size_base") -> nn.Tensor:
   """
-  :return: source with sparse_dim set to out_dim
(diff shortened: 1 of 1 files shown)


albertz:
I don't exactly understand why you need this. Can't you directly do sth like `this_dim.declare_same_as(other_dim)` on returnn-common side?

michelwi:
> Or why are they not the same, if they should be the same?

One use case I could imagine is in the down-sampling + up-sampling approaches where e.g.
```
classes-time = 67
features-time= 67 (can have same dim tag as time-classes)
downsample(features-time) = 23  (factor 3 down-sampling)
upsample(downsample(features-time)) = 69  (factor 3 up-sampling)
```
then using the now wrong placeholder via an `reinterprete_data` layer, the additional 2 frames will be ignored via the sequence mask of `classes-time`

Atticus1806:
We are abusing it right now in our setups, and looking for a „correct“ way to do it but right now this is the way our model does it.

We have an alignment which gives too short lengths so our upsampling is a bit too short so this „ignores“ the last positions of the audio data (for tts)

michelwi:
> This is an example where they are not the same. But then resetting size_base is actually hacky or wrong. [...]
> I assume you just have a softmax + CE loss afterwards [...]  so due to that implementation detail (flattening of the framewise loss), it works.

This is entirely correct and I agree with your assessment. In some cases, where it did not work, I used
* LengthLayer from data, to get expected length
* LengthLayer from upsampled layer, to get the actual length
* EvalLayer to compute the cutoff needed
* SliceNdLayer from upsampled layer to get the exactly needed length

not sure if this is the most efficient method.

albertz:
Ok, so I guess we all agree that instead of using this here, we should rather fix it the "right" way. So, more specifically:

* For `TransposedConvLayer`, implement that it correctly uses the `out_spatial_dims`.
* Implement some `CutOffLayer`. Or, actually, isn't `SliceNdLayer` pretty much that already? You can pass `size` there to e.g. the wanted out spatial dim, or otherwise pass the output of a `LengthLayer` to `size`.

Atticus1806:
so something like?
```python
      align_len = nn.length(dec_lin, axis=align_time_dim)
      target_speech, _ = nn.slice_nd(target_speech, axis=speech_time, start=some_0_tensor, size=align_len, out_spatial_dim=align_time_dim)
```
maybe allow for ints in `slice_nd` `start` which gets broadcast in `batch_dim`?

but I also saw, `slice_nd` takes a dim for size, so maybe this also works?
`target_speech, _ = nn.slice_nd(target_speech, axis=speech_time, start=some_0_tensor, size=align_time_dim, out_spatial_dim=align_time_dim)`

I agree, this would be way nicer than abusing the reinterpret layer.

albertz:
`SliceNdLayer` change: https://github.com/rwth-i6/returnn/pull/1081

albertz:
`TransposedConvLayer` change: https://github.com/rwth-i6/returnn/pull/1083

albertz:
Everything is implemented now.

PR #196, opened 2022-08-09
generate layers align

Atticus1806:
Generate layers for https://github.com/rwth-i6/returnn/pull/1091


Atticus1806:
Test that fails is the same on master, so this should be safe to merge.

albertz (approved):

PR #215, opened 2022-10-05
remove lazy init logic, Linear, Conv, Norm API changes

### LINKED ISSUE #109: Parameter random init outside name ctx does not work

### LINKED ISSUE #212: Lazy init causes unexpected behavior?
I was writing this code:
```python
def test_param_name_deep():
  class _SubNet(nn.Module):
    def __init__(self, out_dim: nn.Dim):
      super().__init__()
      self.linear = nn.Linear(out_dim)

    def __call__(self, x) -> nn.Tensor:
      return self.linear(x)

  class _Net(nn.Module):
    def __init__(self):
      super().__init__()
      self.out_dim = nn.FeatureDim("linear-out", 13)
      self.sub = _SubNet(self.out_dim)
      self.sub2 = _SubNet(self.out_dim)
      self.sub2.linear.weight = self.sub.linear.weight
      self.sub2.linear.bias = self.sub.linear.bias

    def __call__(self, x: nn.Tensor) -> nn.Tensor:
      return self.sub(x) + self.sub2(x)

  net = _Net()
  config, net_dict = dummy_config_net_dict(net)
  pprint(config)
  dummy_run_net(config, net=net)
```
Note the intended param sharing of `sub` an `sub2`.

This code does not produce any error, but it does not share the params! This was somehow unexpected.

The reason is the lazy init of the params when the input dim is not specified, for `nn.Linear` here. (Related: #75)
At this point, `weight` and `bias` in `_Net.__init__` are in fact `None` and this just does nothing.
The params are only set after the first `__call__`.

This however now makes the param sharing somewhat annoying. My code looks now like this:
```python
  class _Net(nn.Module):
    def __init__(self):
      super().__init__()
      self.out_dim = nn.FeatureDim("linear-out", 13)
      self.sub = _SubNet(self.out_dim)
      self.sub2 = _SubNet(self.out_dim)

    def __call__(self, x: nn.Tensor) -> nn.Tensor:
      y = self.sub(x)
      # Can only assign params now due to lazy init...
      self.sub2.linear.initialized = True
      self.sub2.linear.in_dim = x.feature_dim
      self.sub2.linear.weight = self.sub.linear.weight
      self.sub2.linear.bias = self.sub.linear.bias
      return y + self.sub2(x)
```
So I had to move the param sharing logic over to the `__call__`, right after the first `sub(x)` call.
I don't really like this.

So, the question: Should we remove the lazy init logic? Or anything else we should do?
(@JackTemaki @Atticus1806)
(Related: #149)

albertz:
Fix #212.
Fix #109.

And a bunch of semi related other fixes.

This will be rebased and merged, not squashed.

### DIFF
--- a/nn/attention.py
+++ b/nn/attention.py
@@ -35,9 +35,11 @@ class GenericSelfAttention(nn.Module):
   """
   Shared base class for self attention
   """
-  def __init__(self, *, key_dim_total: nn.Dim, value_dim_total: nn.Dim, num_heads: Union[int, nn.Dim],
+  def __init__(self, in_dim: nn.Dim, *,
+               key_dim_total: nn.Dim, value_dim_total: nn.Dim, num_heads: Union[int, nn.Dim],
                att_dropout: float = 0.1):
     super().__init__()
+    self.in_dim = in_dim
     if isinstance(num_heads, int):
       num_heads = nn.SpatialDim("num_heads", num_heads)
     self.key_dim_total = key_dim_total
@@ -47,7 +49,7 @@ class GenericSelfAttention(nn.Module):
     self.num_heads = num_heads
     self.qkv_dim_total = 2 * key_dim_total + value_dim_total
     self.qkv_dim_per_head = 2 * self.key_dim_per_head + self.value_dim_per_head
-    self.qkv = nn.Linear(self.qkv_dim_total)
+    self.qkv = nn.Linear(in_dim, self.qkv_dim_total)
     self.att_dropout = att_dropout
 
   def default_initial_state(self, *, batch_dims: Sequence[nn.Dim]) -> nn.LayerState:
--- a/nn/base.py
+++ b/nn/base.py
@@ -595,7 +595,7 @@ class Parameter(Tensor):
   @initial.setter
   def initial(self, value: Optional[Union[nn.Tensor, RawTensorTypes, nn.init.VarianceScaling]]):
     if isinstance(value, nn.init.VarianceScaling):
-      value = value(self.shape_ordered)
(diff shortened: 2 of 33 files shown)


albertz:
Still WIP but most of it looks ok already.

Most failing tests are because the naming logic changed slightly. But I think it's not optimal yet, and what the tests are checking is right, i.e. in those cases we want the old behavior. I will see how to get that.

albertz:
I just realized, the `deepcopy` logic is even more problematic, as it is implemented currently here. Also see #109 for discussion on that.

Now, I think `nn.Parameter` will not correctly be copied. It will create a copy of `nn.Tensor` and `nn.Parameter`, however, inside the object, the `name_ctx` will not be copied (due to new `NameCtx.__deepcopy__`). This leads to unexpected und buggy behavior.

How to solve this? Again custom `__copy__` and `__deepcopy__` also for `nn.Tensor` and `nn.Parameter`? As mentioned before, `nn.Tensor` is immutable, so it should not do anything. But `nn.Parameter` must get a new copy.

Or change `NameCtx.__deepcopy__`? But this is probably even more difficult, and we anyway need the special logic for the tensors.

Or `Module.__deepcopy__`? But how? But also, it would be unexpected for the user when copying non-module objects (consisting of tensors or so) behave different. So this is also not really an option.

Issue #208, opened 2022-09-14
Change indentation to 4 spaces?


albertz:
It is 2 spaces for consistency to RETURNN. (RETURNN has 2 spaces for historical reasons.)
However, i6_experiments and i6_core use 4 spaces.

Change this here as well? (Maybe also in RETURNN? But that would be maybe a separate discussion. -> https://github.com/rwth-i6/returnn/issues/1156)

albertz:
@JackTemaki @christophmluscher @Atticus1806 I tagged you to get some opinion on this. Maybe also @curufinwe, @michelwi, @mmz33 ?

christophmluscher:
My reasoning would be returnn_common from a code base point of view belongs to returnn and not to Sisyphus / i6_*. IMHO code convention should follow returnn.

In general as a group we should maybe strive for common code formatting rules. But I don't think this is a must.

What do you see as the clear benefit from switching indentation?

albertz:
> What do you see as the clear benefit from switching indentation?

Consistency.

Now in my Sisyphus i6_core/i6_experiments/returnn_common based setup, I basically work on all these three repos. Not so much RETURNN itself anymore, so not sure if your initial statement is really correct.

But as said, maybe we also should change RETURNN itself, but that would be a somewhat separate discussion.

albertz:
> > What do you see as the clear benefit from switching indentation?
> 
> Consistency.

To extend further on this:

Sometimes I even move code from i6_experiments over to returnn_common, or vice versa.

christophmluscher:
> My reasoning would be returnn_common from a code base point of view belongs to returnn and not to Sisyphus / i6_*. IMHO code convention should follow returnn.

so do you agree with this??

> Sometimes I even move code from i6_experiments over to returnn_common, or vice versa.

but from a practical point of view, it seems i6_experiments and returnn_common are closer together?

what would be the main disadvantage in your opinion switching indentation in returnn / returnn_common?
for example we could just run black, so the transition should be fairly straightforward..

albertz:
> > My reasoning would be returnn_common from a code base point of view belongs to returnn and not to Sisyphus / i6_*. IMHO code convention should follow returnn.
> 
> so do you agree with this??

[...]

PR #251, opened 2023-02-02
Apply black, ident by 4 spaces

### LINKED ISSUE #230: Apply black?
Related to 4 space indent (#208) is the question whether we want to have Black applied (as a requirement) to the code, similar as in i6_core or i6_experiments ([i6_core CI test](https://github.com/rwth-i6/i6_core/blob/main/.github/workflows/black.yml)).

120 character limit.

Same issue for RETURNN: https://github.com/rwth-i6/returnn/issues/1158

Let's actually do the discussion in that RETURNN issue, as it probably would be mostly the same here. If you comment here, only comment about anything which would be specific to returnn-common and not apply for RETURNN.

albertz:
Fix #230, #208.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,5 +10,5 @@ tab_width = 4
 trim_trailing_whitespace = true
 
 [*.py]
-indent_size = 2
+indent_size = 4
 indent_style = space
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,22 @@
+name: black
+
+on:
+  push:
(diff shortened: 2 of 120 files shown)


albertz:
Ok, I needed to modify PyCharm to ignore E203, because of this: https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated

albertz:
Now I still get errors like:
```
Pep8CodeStyle: E501: line too long (130 > 120 characters)
```

PR #256, opened 2023-03-13
add nn.stft 

JackTemaki:
I ignored all other layers from signal_processing for now, we can add them gradually when needed.


albertz (changes requested):
`mel_filterbank` should be excluded as well. I would only add low-level functional layers, like `stft`, which wrap some low-level TF function.

Most of the other layers, including `mel_filterbank`, look like they are very simple wrappers around other low-level code. E.g. `mel_filterbank` is just `tf.einsum` (so `nn.dot`) and `tf.tile` (so `nn.tile`) (actually, it looks like `tf.tile` is also not needed here? I did not really look at it so carefully). We don't need to (and should not) wrap such custom layer for that. Rather, we should directly have such `mel_filterbank` implementation, which then calls to `nn.dot` and `nn.tile`.

albertz (approved):

